PR #9, opened 2021-07-25
difftest: rebase xs difftest branch to master

AugustusWillisWang:
Now NEMU master branch can be used as XiangShan difftest reference

BREAKING CHANGE: difftest api has been changed, new difftest framework should be used for XiangShan


sashimi-yzh:
I add a new feature to Makefile yesterday, which supports blacklist to exclude certain source files from compiling according to the configuration of menuconfig.

But now let this branch usable to XiangShan first. I will refactor code later according to the change in PR.

sashimi-yzh (approved):

PR #10, opened 2021-07-26
rv64: fix softfloat-fp

AugustusWillisWang:
Now NEMU, using softfloat, is able to pass `rv64uf` and `rv64ud` in riscv-tests


sashimi-yzh:
Why we need to add a new rule `all-shared` in `resource/softfloat/Makefile`? It seems that you only add a new option `-shared` to gcc, but it is an option for linking, and it should not affect the compling with `-c`.

So do you encounter any errors with `-shared` removed in the softfloat repo?

sashimi-yzh:
It seems that we only need to add `-fPIC` to softfloat when building NEMU as an shared object.

sashimi-yzh:
@AugustusWillisWang Please try my new commit.

AugustusWillisWang:
That's much better, thanks a lot 😂

sashimi-yzh (approved):

PR #15, opened 2021-07-28
chore: fix dependency for softfloat-fp.h

AugustusWillisWang:
It should fix the following compile error:

src/engine/interpreter/softfloat-fp.h:4:10: fatal error:
softfloat.h: No such file or directory


sashimi-yzh:
@AugustusWillisWang I refactor the makefile to fix such compile error. Please try it.

AugustusWillisWang:
No problem, it is working correctly now. @sashimi-yzh

PR #17, opened 2021-08-02
Fix bug in function nemu_large_memcpy.

huxuan0307:
Every bits should be copy from dut. Otherwise, uninitialized ref memory would be random values, which leads to random results when launching difftest.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -229,9 +229,9 @@ config GUIDED_EXEC
   default y
 
 config LARGE_COPY
-  depends on SHARE
+  depends on SHARE && !MEM_RANDOM
   bool "Enable difftest large memory copy optimization"
-  default y
+  default n
 
 config PANIC_ON_UNIMP_CSR
   depends on SHARE


AugustusWillisWang:
> Every bits should be copy from dut. Otherwise, uninitialized ref memory would be random values, which leads to random results when launching difftest.

`nemu_large_memcpy` is intended to avoid copying random (garbage) values to speedup difftest. To avoid random results caused by uninitialized ref memory, you can disable `Enable difftest large memory copy optimization` in menuconfig to use `memcpy` instead of ``nemu_large_memcpy``.

To be more specific:

`make menuconfig`. Enter `Processor difftest reference config`. Disable `Enable difftest large memory copy optimization`. Then recompile NEMU.

huxuan0307:
Thanks for your reply. I enabled "difftest large memory copy optimization" by accident. But are all zero bits garbage? Zero bits in data segment are also garbage? 

I meet the "trait" when I do difftest between my cpu and NEMU. My cpu always produces an exact result while NEMU produces random one. Traced data flow, I find the data copy to NEMU is different from the one in difftest. It's sad to spend almost 2 days on this trait...

May I ask for the docs for NEMU?

AugustusWillisWang:
> But are all zero bits garbage? Zero bits in data segment are also garbage?

I think you are right. Sometimes `nemu_large_memcpy` can be quite dangerous. Workloads used by XiangShan team works correctly on `nemu_large_memcpy` so we enabled it by default. I think it would be much better if we disable `Enable difftest large memory copy optimization` by default in `Kconfig`.

Can you modify the default config here so I can merge your PR?

https://github.com/OpenXiangShan/NEMU/blob/6dcedbced707e50574c266517e3edcf059f443be/Kconfig#L231-L234

AugustusWillisWang:
> May I ask for the docs for NEMU?

Currently there is no doc for master NEMU , NEMU is a work-in-progress and remains in active development. You are welcomed to ask questions and give suggestions on GitHub 😄

poemonsense:
> Thanks for your reply. I enabled "difftest large memory copy optimization" by accident. But are all zero bits garbage? Zero bits in data segment are also garbage?

@huxuan0307 @AugustusWillisWang  We are using `mmap` with flag `MAP_ANONYMOUS`, which initializes all memory bits to zero (refer to https://man7.org/linux/man-pages/man2/mmap.2.html). So there won't be any issues if we only copy the nonzero bits. 

But when memory randomization is enabled via `CONFIG_MEM_RANDOM`, the memory will be initialized to random values, which certainly causes problems. I think it's better to give a warning message when randomization and large_memcpy are both enabled (can be checked in init_mem).

As you discussed before, we still need a default configuration when nemu is running itself. A choice between (1) randomizing the memory (for debugging?); (2) using large_memcpy to avoid allocating unused memory. @sashimi-yzh what's your opinion?

AugustusWillisWang:
https://github.com/OpenXiangShan/NEMU/blob/e59b69e4e271e2bf37b57a902fdbac9a57901c50/Kconfig#L231-L234

@poemonsense We can just disable `large_memcpy` when randomize the initial memory.

sashimi-yzh:
@poemonsense I prefer (1) in `Kconfig` and (2) in `xs_defconfig`, since `large_memcpy` is designed for XiangShan to run large binaries. Running small binaries may not benefit too much from `large_memcpy`.

AugustusWillisWang (approved):

PR #20, opened 2021-08-04
Fix some bugs of riscv64's floating-point

ljwljwljwljw:


ljwljwljwljw:
**fnmadd**
According to the IEEE754 standard, when the rounding mode is not RDN, if the result of the operation is 0, the sign bit of the result is 0, doing fmadd first and then inverting the sign of the result will cause the sign bit of the operation result to be incorrectly set to 1.

When the rounding mode is RDN, the sign of zero result should always be 1, invert the sign bit after doing fmadd will also cause problem.

sashimi-yzh (approved):

AugustusWillisWang (approved):

PR #38, opened 2021-10-20
mmu.asid: add satp's asid length check

Lemover:


Lemover:
nothing changed actually. just add asid check method, now asid can be 0-16, not just only 16.
when change satp reg, flush the soft_tlb

AugustusWillisWang (approved):

PR #46, opened 2021-12-13
intr: fix mtval and stval update condition

poemonsense:
On trap into M-mode, mtval should be updated. Otherwise, stval is updated.


AugustusWillisWang (approved):
paddr r/w is dirty, to be refactored later.

PR #63, opened 2022-04-06
gcpt: skip unnecessary inits when cpt related features are disabled

shinezyy:
fix assertion fault when no output path provided, mentioned in https://github.com/OpenXiangShan/NEMU/issues/62


BigWhiteDog (approved):
seems ok

PR #72, opened 2022-04-22
pmp: when PMP is not supported, still enable pmp r/w, but not check(todo)

Lemover:
When taking checkpoints or restore checkpoints, pmp regs should be stored or restored. But pmp check is not done(still bugs) under OPT mode, so pmp should be disabled under OPT mode. It makes nothing different when running most workloads except that for pmp correctness checking.
So enable pmp read/write even when pmp is disabled for checkpoint taking and restoring.


shinezyy (changes requested):
Please split pmp check and pmp CSR R/W into two macros

PR #108, opened 2022-09-20
fix problems that prohibit mainline linux booting

chenguokai:


huxuan0307:
Bare mode(satp.mode=0) is also allowed!!!

huxuan0307 (approved):

PR #109, opened 2022-10-21
pmp: add config for the number of PMP registers

poemonsense:
This commit adds the config option for the number of PMP registers. The user can configure it via menuconfig.


Lemover (approved):

Lemover:
PMP master

PR #137, opened 2023-08-28
Add raw ELF file load.

yaozhicheng:
Add raw ELF file load.


yaozhicheng:
go back to fix link error

yaozhicheng:
need update

PR #179, opened 2023-09-14
Add TLB to bypass the second stage translation(GPA to HPA)

BL-GS:
The origin 2-stage translation do not utilize TLB but walk the whole translation path, which consume a lot. I add a HOSTVMTLB for the translation from GPA to HPA.
Otherwise, I also add a TLB mapping GVA to PA.
The effect of acceleration depends on how long the VM stays. By test, it accelerates xvisor workload by 300%, and keep almost the same speed under KVM workload.


src/isa/riscv64/system/priv.c
```diff
@@ -801,24 +804,24 @@ static word_t priv_instr(uint32_t op, const rtlreg_t *src) {
           if(cpu.v) longjmp_exception(EX_VI);
           if(cpu.mode == MODE_U) longjmp_exception(EX_II);
           if(!(cpu.mode == MODE_M || (cpu.mode == MODE_S && !cpu.v))) longjmp_exception(EX_II);
-          mmu_tlb_flush(*src);
+          mmu_vmtlb_flush(*src);
```
pxk27, line 807: Read the function of "hfence.vvma" in _the RISC-V Instruction Set Manual
Volume II: Privileged Architecture_

src/isa/riscv64/system/priv.c
```diff
@@ -801,24 +804,24 @@ static word_t priv_instr(uint32_t op, const rtlreg_t *src) {
...
           if(!(cpu.mode == MODE_M || (cpu.mode == MODE_S && !cpu.v && mstatus->tvm == 0))) longjmp_exception(EX_II);
-          mmu_tlb_flush(*src);
+          mmu_vmtlb_flush(*src);
           break;
 #ifdef CONFIG_RV_SVINVAL
         case 0x13: // hinval.vvma
           if(cpu.v) longjmp_exception(EX_VI);
           if(cpu.mode == MODE_U) longjmp_exception(EX_II);
-          mmu_tlb_flush(*src);
+          mmu_vmtlb_flush(*src);
```
pxk27, line 819: Read the function of  "hinval.vvma" in _the RISC-V Instruction Set Manual
Volume II: Privileged Architecture_

src/memory/host-tlb.c
```diff
@@ -53,29 +114,60 @@ void hosttlb_flush(vaddr_t vaddr) {
...
 static word_t hosttlb_read_slowpath(struct Decode *s, vaddr_t vaddr, int len, int type) {
   paddr_t paddr = va2pa(s, vaddr, len, type);
   word_t data = paddr_read(paddr, len, type, cpu.mode, vaddr);
   if (likely(in_pmem(paddr))) {
-    HostTLBEntry *e = type == MEM_TYPE_IFETCH ?
-      &hostxtlb[hosttlb_idx(vaddr)] : &hostrtlb[hosttlb_idx(vaddr)];
-    e->offset = guest_to_host(paddr) - vaddr;
-    e->gvpn = hosttlb_vpn(vaddr);
+    hosttlb_insert(vaddr, paddr, type);
+    hosttlb_insert(vaddr, paddr, type);
```
pxk27, line 170: insert tlb twice?

src/memory/host-tlb.c
```diff
@@ -86,46 +178,43 @@ static void hosttlb_write_slowpath(struct Decode *s, vaddr_t vaddr, int len, wor
...
+    return host_read(dst_ptr, len);
   }
 }
 
 void hosttlb_write(struct Decode *s, vaddr_t vaddr, int len, word_t data) {
-  #ifdef CONFIG_RVH
+#ifdef CONFIG_RVH
   extern bool has_two_stage_translation();
-  if(has_two_stage_translation()){
+  if (has_two_stage_translation()) {
```
pxk27, line 207: The reason why I put two stage translation in this place is to skip tlb translation when NEMU don't have tlb for the second stage. If you add the second stage tlb to NEMU, you should change these codes.

src/memory/host-tlb.c
BL-GS: Maybe the problem about merging, I have fixed it.

src/memory/host-tlb.c
BL-GS: But I only add the TLB for the second stage, which means it still need to skip previous host tlb lookup(va -> pa). 

P.S.
Maybe I can add a TLB mapping GVA to GPA. However, this cannot avoid adding judgement here, and asks page table walker(ptw) to be more granular. I used to do that, but it seems quiet difficult to keep it from a number of macros. Maybe when everything has been done, a code review can be applied.

PR #191, opened 2023-09-22
vaddr: check memory alignment when mmu mode is translate

xyyy1420:


poemonsense (approved):

poemonsense:
I think misaligned exceptions happen when the memory access address is not aligned to the request length. For example, load word should have aligned addresses w.r.t 4 bytes, such as `0x0`, `0x4`, `0x8, `0xc`.

I think we should not shift the address with a page offset?

poemonsense:
@pxk27 would you mind take a look at this CI failure as well? After fixing the address misalignment checker, H ext CI fails.

xyyy1420:
> I think misaligned exceptions happen when the memory access address is not aligned to the request length. For example, load word should have aligned addresses w.r.t 4 bytes, such as , , 0xc`.`0x0``0x4``0x8, `
> 
> I think we should not shift the address with a page offset?

Yes, we shouldn't use page offsets to move addresses, original misalignment check is ok. I was confusing memory alignment with page alignment.

pxk27 (approved):

PR #197, opened 2023-10-17
Fix vcompute logics.

Siudya:
#According to 5.4 of RVV 1.0, vd should not be updated when vstart >= vl.


src/isa/riscv64/instr/rvv/vcompute_impl.c
```diff
@@ -95,6 +95,14 @@ static inline void update_vcsr() {
 }
 
 void arthimetic_instr(int opcode, int is_signed, int widening, int narrow, int dest_mask, Decode *s) {
+  if(vstart->val >= vl->val) {
```
poemonsense, line 98: It seems the duplicated code can be wrapped as a function, which is shared by different functions

PR #207, opened 2023-11-03
Fix some bugs about checkpoint

BL-GS:
1. When generating and executing checkpoints according to the tutorials or README.md, it **may** incur confusing errors. I found that **NEMU only restores a part of `gcpt.bin`**.

> What is incredible is that scripts(restore.sh) uses different shell commands from those in tutorials and thus bypasses this question... I may stuck here for longer if I haven't look at your CI workflow.
2. The directory name become negative number when `inst_count` excesses the limitation of `int`. I fixed it.

What I haven't solved is that NEMU will generate `csrrs x10, 0x0c1, x0` instruction, which is not supported by XiangShan. I found it resulted from following codes:

`csr.h`
```c
#define CSRS(f) \
...
  f(mtime      , 0xc01) \
```
`resource/gcpt_restore/src/restore.S`
```asm
restore_csr_vector:
  # set fs
  li t0, MSTATUS_FS
  csrs  CSR_MSTATUS, t0

  li t0, CSR_CPT_ADDR
  CSRS(CSRS_RESTORE)
  HCSRS(CSRS_RESTORE)
```
To fix this may acquire more discussion.

### DIFF
--- a/include/checkpoint/path_manager.h
+++ b/include/checkpoint/path_manager.h
@@ -32,7 +32,7 @@ class PathManager
     std::string configName;
     std::string workloadName;
 
-    int cptID;
+    uint64_t cptID;
 
     std::string workloadPath;
     fs::path outputPath;
@@ -43,7 +43,7 @@ class PathManager
 
     void incCptID();
 
-    int getCptID() const {return cptID;}
+    uint64_t getCptID() const {return cptID;}
 
     std::string getOutputPath() const;
 
--- a/include/checkpoint/serializer.h
+++ b/include/checkpoint/serializer.h
@@ -48,7 +48,7 @@ class Serializer
 
     uint64_t intervalSize{10 * 1000 * 1000};
 
-    int cptID;
+    uint64_t cptID;
     std::string weightIndicator;
 
     const uint32_t IntRegStartAddr;
--- a/src/checkpoint/path_manager.cpp
+++ b/src/checkpoint/path_manager.cpp
@@ -51,7 +51,7 @@ void PathManager::init() {
   //    cptID = cpt_id;
   //  }
 
-  Log("Cpt id: %i", cptID);
+  Log("Cpt id: %li", cptID);
 
   workloadPath = statsBaseDir + "/" + configName + "/" + workloadName + "/";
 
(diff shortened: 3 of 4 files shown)


xyyy1420:
1. That's my fault, thanks for fixing. By the way, restore.sh using --cpt-restore solves(with some limitations) the problem caused by the difference between the older gcpt.bin and the new gcpt.bin, rather than bypass it.

BL-GS:
As for the third change, my prior explanation might be a little confusing. Let me give an example(without the change):
1. At first, we generate checkpoints by NEMU, which emulates user-level timer(0xc01) successfully and set `SBI_HART_EXT_ZICNTR` bit in the variable `hfeature`.
```c
  /* Detect if hart supports time CSR */
  csr_read_allowed(CSR_TIME, (unsigned long)&trap);
  if (!trap.cause)
	  __sbi_hart_update_extension(hfeatures,
				  SBI_HART_EXT_ZICNTR, true);
```
2. We run checkpoint on Xiangshan, and it restores the variable `hfeature` in memory.
3. When application executes instruction **rdtime** and tries to read user-level timer(0xc01), it raises illegal instruction exception, enters M-mode and awakes OpenSBI to handle the exception(`opensbi/lib/sbi/sbi_illegal_insn.c`).
4. OpenSBI tries to emulate the behavior. However, this is what it does:
```c
// opensbi/lib/sbi/sbi_timer.c
if (sbi_hart_has_extension(scratch, SBI_HART_EXT_ZICNTR))
	get_time_val = get_ticks;
```
where the function `get_ticks` is:
```c
// opensbi/lib/sbi/sbi_timer.c
#if __riscv_xlen == 32
static u64 get_ticks(void)
{
	u32 lo, hi, tmp;
	__asm__ __volatile__("1:\n"
			     "rdtimeh %0\n"
			     "rdtime %1\n"
			     "rdtimeh %2\n"
			     "bne %0, %2, 1b"
			     : "=&r"(hi), "=&r"(lo), "=&r"(tmp));
	return ((u64)hi << 32) | lo;
}
#else
static u64 get_ticks(void)
{
	unsigned long n;

	__asm__ __volatile__("rdtime %0" : "=r"(n));
	return n;
}
#endif
```
Yeah, it executes the instruction `rdtime` again in M-mode. It will raise illegal instruction exception, but OpenSBI does not allow illegal instruction exception in M-mode(Even if it did, the program would be stuck in an infinite loop):
```c
// opensbi/lib/sbi/sbi_illegal_insn.c
if (prev_mode == PRV_M) {
	sbi_printf("%s: Failed to access CSR %#x from M-mode",
		__func__, csr_num);
	return SBI_EFAIL;
}
```

xyyy1420 (commented):
seems ok

xyyy1420:
I apologize for taking so long to reply to this, can you fix the conflict?

PR #231, opened 2023-12-11
configs: Increase dual-xs-ref memory size

good-circle:


poemonsense:
#319 do the same thing.

PR #237, opened 2023-12-21
interrupt: shield spike mtip stip utip

xiaokamikami:
Prevent the spike from repeatedly triggering the time interrupt

### DIFF
--- a/src/isa/riscv64/difftest/dut.c
+++ b/src/isa/riscv64/difftest/dut.c
@@ -29,6 +29,9 @@
 
 bool isa_difftest_checkregs(CPU_state *ref_r, vaddr_t pc) {
   csr_prepare();
+#ifdef CONFIG_DIFFTEST_REF_SPIKE
+  cpu.mip &= 0xffffff4f; // ignore difftest for mip
+#endif
   if(cpu.mip != ref_r->mip) ref_r->mip = cpu.mip; // ignore difftest for mip
   if (memcmp(&cpu.gpr[1], &ref_r->gpr[1], DIFFTEST_REG_SIZE - sizeof(cpu.gpr[0]))) {
     int i;


poemonsense:
`ref.c` is NEMU as a REF. If you are co-simulating NEMU with Spike, NEMU is the DUT?

poemonsense (changes requested):
please explain why to change `ref.c`

xiaokamikami:
> please explain why to change `ref.c`

https://github.com/OpenXiangShan/riscv-isa-sim/pull/10
Same as this, except the processing is moved from spike to nemu, because copy tip causes repeated interrupts that trigger spike

poemonsense:
> > please explain why to change `ref.c`
> 
> [OpenXiangShan/riscv-isa-sim#10](https://github.com/OpenXiangShan/riscv-isa-sim/pull/10) Same as this, except the processing is moved from spike to nemu, because copy tip causes repeated interrupts that trigger spike

See https://github.com/OpenXiangShan/difftest/issues/30.

To co-simulate a DUT and a REF, only the reg part of mip of **DUT** should be copied to REF. It does not state that REF should not copy the mtip stip utip bits.

If you are to fix the NEMU (as a DUT) co-simulation with Spike (as a REF), you should fix the `dut.c` of NEMU instead of `ref.c` of NEMU. `ref.c` is for NEMU (as a REF) to simulate with other designs.

poemonsense:
Same, if you are to fix the Spike (as a DUT) for co-simulation with NEMU (as a REF), you should fix the DUT code of Spike.

xiaokamikami:
I updated the code to dut.c

src/isa/riscv64/difftest/ref.c
```diff
@@ -59,7 +59,9 @@ void csr_prepare() {
   cpu.sepc    = sepc->val;
 
   cpu.satp     = satp->val;
+
```
poemonsense, line 62: please avoid these unnecessary space changes

xiaokamikami: ok

src/isa/riscv64/difftest/dut.c
```diff
@@ -29,7 +29,10 @@
 
 bool isa_difftest_checkregs(CPU_state *ref_r, vaddr_t pc) {
   csr_prepare();
-  if(cpu.mip != ref_r->mip) ref_r->mip = cpu.mip; // ignore difftest for mip
+#ifdef CONFIG_DIFFTEST_REF_SPIKE
+  cpu.mip &= 0xffffff4f; // ignore difftest for mip
```
poemonsense, line 33: Should we mask the bits off for `ref_r->mip` instead of zeroing these bits for cpu?

Such as `ref_r->mip &= the_mask`? Unless the cpu states will be changed by difftest. Is this expected?

src/isa/riscv64/difftest/ref.c
```diff
@@ -60,6 +60,7 @@ void csr_prepare() {
 
   cpu.satp     = satp->val;
   cpu.mip      = mip->val;
+
```
poemonsense, line 63: one more blank line left here. Please check "Files changed" before submitting PRs. Or check `git diff` before push

src/isa/riscv64/difftest/dut.c
poemonsense: I still doubt whether we should change the `cpu.mip` instead of copying partial of it. But if you think this's ok, I'm now approving this

poemonsense (approved):

poemonsense:
I still doubt whether we should change the cpu.mip instead of copying partial of it. But if you think this's ok, I'm now approving this

PR #244, opened 2024-01-02
CI: add vectore test

xiaokamikami:
Add vectore's test for CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master , rv64v]
   pull_request:
-    branches: [ master ]
+    branches: [ master , rv64v]
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - name: Setup env
         run: |
           echo "NEMU_HOME=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          git checkout master
```
poemonsense, line 21: no need for this. CI will automatically checkout to the HEAD commit to run tests

xiaokamikami: ok

.github/workflows/ci.yml
```diff
@@ -128,3 +129,28 @@ jobs:
...
+    name: nemu - V-extension
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup env
+        run: |
+          echo "NEMU_HOME=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "TEST_HOME=/nfs/home/share/ci-workloads/V-extension-tests" >> $GITHUB_ENV
+          git fetch
+          git checkout rv64v
```
poemonsense, line 143: To test rv64v branch only, should use `branches: [ rv64v ]` to specify the condition of CI and this CI file is updated only on the rv64v branch. No need to merge into master.

What's the target of this PR? To test RVV on rv64v branch?

If so, you can modify this PR over the rv64v branch, and set the CI condition. Then, github will automatically run CI on the rv64v branch.

.github/workflows/ci.yml
```diff
@@ -128,3 +128,28 @@ jobs:
...
+      - name: Build NEMU with V extension as difftest ref
+        run: |
+          make riscv64-rvv-diff-spike_defconfig
+          make -j
+          make clean-all
+      - name: Run Vector-hmmer with Spike DiffTest
+        run: |
+          make riscv64-rvv-diff-spike_defconfig
+          echo "#define CONFIG_DETERMINISTIC 1" >> ./include/generated/autoconf.h
```
poemonsense, line 152: is this `CONFIG_DETERMINISTIC ` required?

.github/workflows/ci.yml
xiaokamikami: Yes, we only want to do this test on rv64v for now, we will need to add it to master later

PR #257, opened 2024-01-25
CI: add hammer as V extension test (#244)

xiaokamikami:


Ziyue-Zhang (approved):

poemonsense:
title should be hmmer instead of hammer?

Ziyue-Zhang:
> title should be hmmer instead of hammer?

sorry, I misspelled it

PR #258, opened 2024-01-25
vector-diff: vector store not diff enough times

xiaokamikami:
Fixed the case that store diff does not diff enough times under V vector memory access

### DIFF
--- a/src/memory/paddr.c
+++ b/src/memory/paddr.c
@@ -341,6 +341,12 @@ int check_store_commit(uint64_t *addr, uint64_t *data, uint8_t *mask) {
   return result;
 }
 
+inline uint64_t store_read_step() {
+  if (tail >= head) 
+    return tail - head;
+  else 
+    return (CONFIG_DIFFTEST_STORE_QUEUE_SIZE - head + tail);
+}
 #endif
 
 char *mem_dump_file = NULL;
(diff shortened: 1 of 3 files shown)


src/memory/paddr.c
```diff
@@ -91,7 +91,6 @@ static inline void pmem_write(paddr_t addr, int len, word_t data) {
 #ifdef CONFIG_DIFFTEST_STORE_COMMIT
   store_commit_queue_push(addr, data, len);
 #endif
-
```
poemonsense, line 94: No need to remove the blank line here?

poemonsense (commented):
I'm not quite familiar with the vector diff here.

Does NEMU push multiple stores to the queue for vector store? In such case, do we need to pop all entries for store difftest?

Probably we can only change the `DIFFTEST_STORE_QUEUE_SIZE` when vector is enabled. For example, change the config for the corresponding config files?

Overall, this PR seems good to me. Does anyone else implement the vector diff? I'm not quite familiar with the code here.

poemonsense (approved):

xiaokamikami:
> I'm not quite familiar with the vector diff here.
> 
> Does NEMU push multiple stores to the queue for vector store? In such case, do we need to pop all entries for store difftest?
> 
> Probably we can only change the `DIFFTEST_STORE_QUEUE_SIZE` when vector is enabled. For example, change the config for the corresponding config files?
> 
> Overall, this PR seems good to me. Does anyone else implement the vector diff? I'm not quite familiar with the code here.

Ok, let me optimize it

PR #260, opened 2024-01-26
doc: update doc for NEMU's usage in Xiangshan workflow

shinezyy:


shinezyy:
WIP, plz don't merge

README.md
```diff
@@ -1,29 +1,34 @@
...
+* CPU core with support of most common ISAs
   * x86
     * real mode is not supported
     * x87 floating point instructions are not supported
   * mips32
     * CP1 floating point instructions are not supported
   * riscv32
     * only RV32IM
+  * riscv64
+    * rv64gcbh currently
```
poemonsense, line 26: Also support K extension

README.md
```diff
@@ -32,59 +37,224 @@ The main features of NEMU include
...
+## FAQ
+
+### Why cannot produce a checkpoint in M-mode?
+
+Read the source code of [GCPT restorer](https://github.com/OpenXiangShan/NEMU/blob/master/resource/gcpt_restore/src/restore.S)
+
+Because we restore checkpoint in M mode, and the PC of returning to user mode is stored in EPC register.
+This recovery method will break the architecture state (EPC) if the checkpoint is produced in M mode.
+In contrast, if the checkpoint is produced in S mode or U mode,
+the return process is just like a normal interrupt return, which will not break the architecture state.
```
poemonsense, line 219: like a normal trap return

trap is more suitable here for both exceptions and interrupts

README.md
```diff
@@ -32,59 +37,224 @@ The main features of NEMU include
...
+```
+
+### Use NEMU as reference design
+
+#### Build reference.so
+
+To build NEMU as reference design, run
+``` bash
+make xxx-ref_defconfig
+make menuconfig  # save
```
poemonsense, line 157: If I remember it correctly, `make menuconfig` is only required at the first time when NEMU is downloaded (before make any_defconfig).

After that, only make xxx_defconfig is needed to safely build NEMU.

poemonsense (approved):
Also considering replacing all "Xiangshan" with "XiangShan"?

PR #262, opened 2024-01-29
load_gz: one MEMORY_REGION_ANALYSIS use fast load

xiaokamikami:
The speed of loading a single gz file can be improved from 24 seconds to around 17 seconds


shinezyy:
The reaseon we do zero check before write is to keep the page untouched and take advantage of Linux COW to reduce memory usage. If neccesary, you can add a new option to load without checking and turn on zero checking by default.

PR #265, opened 2024-02-12
Difftest: sperate basic store log from lightqs

klin02:
Previously logic of store log is mixed with lightqs. Now we seperate basic store log, so difftest can take efficient snapshot of each squashed commit, and recover memory if this squashed comparation errors. It allows us to replay unsquashed commit for better debuggability.

We enable STORE_LOG for default ref config. It will do nothing unless dynamic_config.enable_store_log is set to true by difftest.


include/common.h
```diff
@@ -43,6 +43,9 @@ typedef uint16_t ioaddr_t;
 struct DynamicConfig {
   bool ignore_illegal_mem_access;
   bool debug_difftest;
+#ifdef CONFIG_STORE_LOG
```
poemonsense, line 46: Probably we should not put `bool enable_store_log;` under a macro for this struct here.

When multiple macros are included, we cannot determine their positions/offsets in the struct from external code. Different configs may affect this struct, and then difftest may not see the same struct type (to update the config here).

It looks better if `enable_store_log` is always defined but not used when the macro is not defined

poemonsense (commented):
I know little about LightQS.

Is the store_log for store requests to memory? Previously we also have a store request buffer in paddr.c. Do they have the same functionalities?

klin02:
> I know little about LightQS.
> 
> Is the store_log for store requests to memory? Previously we also have a store request buffer in paddr.c. Do they have the same functionalities?

Store log is used to store the original data in memory before pmem_write. Therefore, we record the original data sequence since the snapshot point, and we can restore the memory state of the snapshot point in reverse order.

include/common.h
klin02: Sure. I will fix it

poemonsense:
> > I know little about LightQS.
> > Is the store_log for store requests to memory? Previously we also have a store request buffer in paddr.c. Do they have the same functionalities?
> 
> Store log is used to store the original data in memory before pmem_write. Therefore, we record the original data sequence since the snapshot point, and we can restore the memory state of the snapshot point in reverse order.

How do you restore the memory of DUT? Also record the store requests?

I still think using fork would be better

poemonsense:
> > > > I know little about LightQS.
> > > > Is the store_log for store requests to memory? Previously we also have a store request buffer in paddr.c. Do they have the same functionalities?
> > > 
> > > 
> > > Store log is used to store the original data in memory before pmem_write. Therefore, we record the original data sequence since the snapshot point, and we can restore the memory state of the snapshot point in reverse order.
> > 
> > 
> > How do you restore the memory of DUT? Also record the store requests?
> > I still think using fork would be better
> 
> It seems we don't need to restore memory of DUT? When using Squash Replay, we take a snapshot of the REF before every squashed Diffstate handled, but restore the REF state only if an error occurs, notifying the DUT to transmit the unsquashed DiffState (already generated and stored), terminating the simulation at one of corresponding unsquashed State.
> 
> Overall, only when an squashed DiffState is error, we will notify DUT to transmit stored unsquashed DiffState, and restore REF before squashed State, to process corresponding unsquashed state.

Make sense. Only REF states are required to be reset.

poemonsense (approved):

PR #266, opened 2024-02-19
support using zstd decompress checkpoint

xyyy1420:


Makefile
```diff
@@ -28,6 +28,7 @@ remove_quote = $(patsubst "%",%,$(1))
 
 ISA    ?= $(if $(CONFIG_ISA),$(call remove_quote,$(CONFIG_ISA)),x86)
 CFLAGS += -D__ISA__=$(ISA)
+CFLAGS += -lzstd
```
poemonsense, line 31: Only include the header when CONFIG_MEM_COMPRESS is enabled?

Makefile
```diff
@@ -28,6 +28,7 @@ remove_quote = $(patsubst "%",%,$(1))
 
 ISA    ?= $(if $(CONFIG_ISA),$(call remove_quote,$(CONFIG_ISA)),x86)
 CFLAGS += -D__ISA__=$(ISA)
+
```
poemonsense, line 31: Unused empty line

poemonsense (commented):
LGTM. Let zyy have a look

src/monitor/image_loader.c
```diff
@@ -63,6 +65,131 @@ long load_gz_img(const char *filename) {
...
+      free(decompress_file_buffer);
+
+      return -1;
+    }
+
+    if (output.pos == 0) {
+      break;
+    }
+
+    for (uint32_t x = 0; x < output.pos; x++) {
```
shinezyy, line 165: Can we use uint64_t here? U64 might be faster

shinezyy (changes requested):

shinezyy (approved):

PR #281, opened 2024-03-15
ref: remove unnecessary compilation links

xiaokamikami:
When compiling as ref, you don't need to enable checkpoint functionality, which saves you 50kb of compile size and avoids some cases where the host doesn't have the relevant library files to run
在作为ref编译时不需要开启checkpoint的相关功能，这样可以减少50kb的编译大小，并且避免一些因为宿主机没有相关库文件造成无法运行的情况

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,10 +57,10 @@ SRCS-y += $(shell find $(DIRS-y) -name "*.c")
 
 SRCS = $(SRCS-y)
 
-DIRS-y += src/profiling # profiling.c
+DIRS-y += src/profiling
 ifndef CONFIG_SHARE
 DIRS-cpp = src/checkpoint src/base src/iostream3 src/memory src/profiling
-DIRS-y += src/checkpoint # profiling.c
+DIRS-y += src/checkpoint
 
 XSRCS = $(shell find $(DIRS-cpp) -name "*.cpp")
 endif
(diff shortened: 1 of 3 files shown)


include/checkpoint/cpt_env.h
```diff
@@ -13,6 +13,7 @@
 * See the Mulan PSL v2 for more details.
 ***************************************************************************************/
 
+#ifndef CONFIG_SHARE
```
poemonsense, line 16: If we remove checkpoint c files in the Makefile, do we need to change these files? If we remove them in Makefile, they should not be compiled?

xiaokamikami: It's true that I don't have to change the file definition in checkpoint anymore

Makefile
```diff
@@ -57,10 +57,13 @@ SRCS-y += $(shell find $(DIRS-y) -name "*.c")
 
 SRCS = $(SRCS-y)
 
+DIRS-y += src/profiling # profiling.c
```
poemonsense, line 60: Let's remove the unused comment ` # profiling.c`. It seems useless

poemonsense: Line 63 also has this useless comment

xyyy1420 (commented):
seems good

xyyy1420 (approved):

PR #284, opened 2024-03-20
fix: fix sv39 mtval

xiaokamikami:
In xiangshan, the virtual address is trimmed to save the chip area. When SAM LAM exception is triggered, only the address with a large size of 39 bits will be written, so certain changes need to be made when xs ref is used


poemonsense (commented):
I doubt this is the case and believe XS should sign-extend 39bits?

poemonsense (commented):
See https://www.geeksforgeeks.org/sign-extend-a-nine-bit-number-in-c/ for how to sign-extend a number

poemonsense (approved):

PR #308, opened 2024-04-11
memory: Allow DUT to provide pmem for NEMU ref

shinezyy:
When restoring from checkpoints, most memory contents of DUT and reference are the same. Therefore, creating a COW mirror memory in DUT side as the memory of reference design can save memory, especially when using multi-core difftest.


poemonsense (approved):
Both `set_pmem` and `init_mem` may call `allocate_memory_with_mmap`. Will there be any conflicts?

Others look good to me.

PR #315, opened 2024-04-24
clint: use physical uptime for real-time timer

ez4yunfeng2:
修改非 CONFIG_DETERMINISTIC 情况下 mtime计算方法
mtime = 运行时间(微秒) / 每周期微秒数

CLOCK_FREQ 设置为 1000000ul 与 opensbi 中 system.dts 对齐

### DIFF
--- a/src/isa/riscv64/clint.c
+++ b/src/isa/riscv64/clint.c
@@ -20,7 +20,8 @@
 
 #define CLINT_MTIMECMP (0x4000 / sizeof(clint_base[0]))
 #define CLINT_MTIME    (0xBFF8 / sizeof(clint_base[0]))
-#define TIMEBASE 10000000ul
+#define CLOCK_FREQ 1000000ul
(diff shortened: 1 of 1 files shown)


xiaokamikami:
You removed TIMEBASE so that CONFIG_DETERMINISTIC can no longer be enabled, please add it back

in addition
This solves the problem of suspended animation (the system misperceives the thread timeout) when running liunx workload when the machine emulation speed is too slow . @poemonsense

poemonsense:
I don't know much about the CLINT in NEMU. But it seems OK to me. This will make time the same as the actual physical time instead of how long simulation has lasted?

Need to fix the compilation errors when `CONFIG_DETERMINISTIC` is set as suggested

ez4yunfeng2:
Fix compilation error when setting CONFIG_DETERMINISTIC，reset the variable name CLOCK_FREQ to TIMEBASE. @poemonsense

poemonsense (approved):

PR #319, opened 2024-04-28
config: Increase the memory capacity and add RVV in the dual config

xiaokamikami:
Since the V extension implementation is enabled by default in Xiangshan, RVV is enabled by default. 
At the same time, in order to support the operation of checkpoint on multi-core, the memory and single-core config are set to the same configuration (8G MMAP).

PR #321, opened 2024-04-30
memory: mmap difftest pmem_base for dual-core difftest

klin02:
Previous we use MAP_FIXED for single-core, prevent allocating memory repeatedly for each initialization.

For dual-core, we also want to use mmap for MEMORY_SIZE config. To avoid different core conflict with same PMEM_BASE, we apply different pmem_base according to hartid.

### DIFF
--- a/include/memory/paddr.h
+++ b/include/memory/paddr.h
@@ -20,6 +20,7 @@
 #include <common.h>
 
 extern unsigned long MEMORY_SIZE;
+extern unsigned int PMEM_HARTID;
 
 #ifdef CONFIG_MODE_USER
 #define CONFIG_MBASE 0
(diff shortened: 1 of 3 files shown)


shinezyy:
Is map_fixed still necessary? @poemonsense

poemonsense:
> Is map_fixed still necessary? @poemonsense

Yes. NEMU does not provide the API to release (`munmap`) mapped memories. Thus, to avoid resource leaks, MAP_FIXED is required to allow the nemu-so to be initialized multiple times when running multiple workloads

poemonsense (approved):

poemonsense:
Update the default dual  config as well?

klin02:
The dual config has been updated at #319 ?

https://github.com/OpenXiangShan/NEMU/blob/6720b33b87c1b0edd065bd5603b96be24f8f049d/configs/riscv64-dual-xs-ref_defconfig#L91

PR #326, opened 2024-05-11
system: remove the extra printf

poemonsense:


poemonsense:
Introduced in https://github.com/OpenXiangShan/NEMU/pull/216

PR #327, opened 2024-05-12
Add c.ebreak instruction

fawangZ:


src/isa/riscv64/instr/rvc/exec.h
```diff
@@ -67,6 +67,14 @@ def_EHelper(c_jalr) {
 #endif
 }
 
+#ifdef CONFIG_SHARE
```
poemonsense, line 70: Why CONFIG_SHARE?

fawangZ: Just for reference model at first, now this PR is modified.

src/isa/riscv64/instr/rvc/exec.h
```diff
@@ -67,6 +67,18 @@ def_EHelper(c_jalr) {
 #endif
 }
 
+def_EHelper(c_ebreak) {
+#ifdef CONFIG_RV_DEBUG
+  rtl_trap(s, s->pc, 3);
+  rtl_mv(s, s0, t0);
```
poemonsense, line 73: why `rtl_mv(s, s0, t0);` and `rtl_priv_jr(s, s0);`?

From isa/riscv64/instr/priv/system.c, I see:

```c
#ifdef CONFIG_RV_DEBUG
    } else if (id == 1) { // ebreak
      rtl_trap(s, s->pc, 3);
      rtl_mv(s, jpc, t0);
#elif defined(CONFIG_EBREAK_AS_TRAP)
...
#endif
```

So should we only do the `rtl_mv(s, jpc, t0);` here?

fawangZ: `rtl_priv_jr(s, s0);` is executed by ebreak in src/isa/riscv64/instr/priv/exec.h:21, I think c_ebreak should do the same.

poemonsense: I see. Then should we move ` rtl_priv_jr(s, s0);` to `#ifdef CONFIG_RV_DEBUG`? This line isunused in CONFIG_EBBREAK_AS_TRAP which contains a longjmp

fawangZ: Yes, this is more intuitive in code, I will fix it.

poemonsense:
Since other reviewers do not respond to this PR, I'm approving and merging this now as this PR looks good to me. If there're future issues, we will then fix them in future PRs.

PR #129, opened 2023-07-04
RV64：Support Hypervisor Extention Version 1.0

pxk27:
This pull request add RVH version 1.0 support for NEMU

* Support new privilege modes
* Support new hypervisor and virtual supervisor CSRs
* Support new hypervisor instructions
* Modify some machine-level CSRs
* Support two-stage address translation
* Support new traps

PR #337, opened 2024-05-14
csr: fix the read/write masks of sstatus

poemonsense:
sstatus is a masked register of mstatus. Its read and write masks are partial masks of the mstatus masks.

This commit fixes the wmask for sstatus.fs, which should depend on the existence of FPU. We also clean up some duplicated code.


poemonsense:
https://github.com/OpenXiangShan/NEMU/pull/129 copies the `SSTATUS_RMASK` code to ref.c. This PR removes it since this macro has already been defined in csr.h

poemonsense:
https://github.com/OpenXiangShan/NEMU/pull/335 fix the wmask of mstatus.fs. However, since sstatus masks are previously independent of mstatus masks, mask of sstatus.fs is still wrong.

PR #339, opened 2024-05-15
priv: allow writable mstatus.fs for software FPU emulation

poemonsense:
This commit adds a config to allow writable mstatus.fs when FPU is disabled via CONFIG_FPU_NONE.

We made mistakes on the write-only zero features of mstatus.fs when FPU is disabled. According to the RISC-V spec, an implementation may choose not to have the FS field be read-only zero to enable emulation of the floating-point extension for S- and U- modes via invisible traps into M-mode. In such cases, the FPU will be emulated by M-mode firmware.


poemonsense:
https://github.com/OpenXiangShan/NEMU/pull/337 brings kernel panic when the nutshell config boots Linux. We expect the fs to be not read-only zero in most cases to enable software emulation of FPU.

poemonsense:
This PR partially reverts https://github.com/OpenXiangShan/NEMU/pull/335

poemonsense:
This PR also updates the default configs for NutShell with Zicond extensions being disabled. It is by default enabled in the config file.

poemonsense:
> Whether mstatus.fs read-only zero or not is a RISC-V related feature. I think it's not proper to put this config in the main menu.

Though I believe software FPU emulation is a general option for all architectures, currently this option only affects the RISC-V architecture. Despite the disagreement with your opinion, it's ok to reduce the scope of the newly added option to RISC-V.

I've updated the commit.

cebarobot (approved):
LGTM. It's better to put two FS realated config together.

PR #340, opened 2024-05-16
rv64, csr: ensure mvenderid, marchid, mimpid are always implemented

cebarobot:
The RISC-V privileged ISA specifies that "mvendorid, marchid, and mimpid must be readable in any implementation." Therefore, there is no need to retain a switch to control their implementation.

This patch remove RV_ARCH_CSRS from Kconfig, and modify the defconfigs correspondingly. Also, this patch remove an improper mux when initialize marchid.


poemonsense (commented):
These CSRs do not exist in RISC-V priv 1.11. They are introduced in priv 1.12. We can remove this configuration because we will not maintain any 1.11 designs.

The MUX seems designed for difftest between NEMU and Spike. I think it should be there.

But I cannot confirm this @xiaokamikami are you using spike to diff nemu? I don't remember who is doing this.

poemonsense:
I'm not sure which config he is using for difftest. To avoid breaking others workflow, let @pxk27 have a review

Others look good to me

pxk27:
I think there is no problem with these changes

pxk27 (approved):

PR #341, opened 2024-05-16
csr: rewrite csr.h to improve readability & maintainability

cebarobot:
As more and more extentions of RISC-V have been added into NEMU, the CSR part becomes completely unreadable and unmaintainable. This patch aims to enhance readability & maintainability.

In this patch, the order of CSRs and the name of each CSR part follows the RISC-V privileged ISA. All CSR mappings merge into one mapping marco "CSRS". The form of switches, which control CSR existence by specific extension, are unifed.

This patch add these CSRs according to ISA: cycle, time, instret, hpmcounter3~31, senvcfg, scountinhibit, hcontext, mconfigptr, mscontext. These CSRs are still lack of functionality, and they may not be actually existed due to lack of according CONFIG defines. Also, fix wrong name of CSR time (0xC01).

This patch is part of a series aimed at enhancing the functionality, readability, and maintainability of the CSR component.

### DIFF
--- a/src/isa/riscv64/local-include/csr.h
+++ b/src/isa/riscv64/local-include/csr.h
@@ -19,134 +19,305 @@
 
 #include <common.h>
 
-#ifdef CONFIG_RV_SVINVAL
-#define CUSTOM_CSR(f) \
-  f(srnctl     , 0x5c4)
-#else
-#define CUSTOM_CSR(f)
-#endif
+/**
+ * Mapping between CSR name and addr
+ * 
+ * This part follows the order of "CSR Listing" section in Privileged ISA Manual.
+*/
 
-// Debug Mode ISA CSRs in Sdext ISA extension
-#ifdef CONFIG_RVSDEXT
-  #define CORE_DEBUG_CSRS(f) \
-  f(dcsr       , 0x7b0) f(dpc        , 0x7b1) f(dscratch0  , 0x7b2) f(dscratch1  , 0x7b3) \
+/* Unprivileged CSR */
+/** Unprivileged Floating-Point CSRs **/
+#define CSRS_UNPRIV_FLOAT(f) \
+  f(fflags       , 0x001) f(frm        , 0x002) f(fcsr       , 0x003) \
 
-#else
-  #define CORE_DEBUG_CSRS(f)
-#endif // CONFIG_RVSDEXT
+/** Unprivileged Counter/Timers **/
+#ifdef CONFIG_RV_Zicntr
+  #define CSRS_UNPRIV_CNTR(f) \
+    f(cycle      , 0xC00) f(time       , 0xC01) f(instret    , 0xC02)
+#else // CONFIG_RV_Zicntr
+  #define CSRS_UNPRIV_CNTR(f)
+#endif // CONFIG_RV_Zicntr
 
-// Trigger CSRs in Sdtrig ISA extension
+#ifdef CONFIG_RV_Zihpm
+  #define CSRS_UNPRIV_HPMCOUNTER(f) \
+    f(hpmcounter3    , 0xC03) \
+    f(hpmcounter4    , 0xC04) f(hpmcounter5    , 0xC05) f(hpmcounter6    , 0xC06) f(hpmcounter7    , 0xC07) \
+    f(hpmcounter8    , 0xC08) f(hpmcounter9    , 0xC09) f(hpmcounter10   , 0xC0A) f(hpmcounter11   , 0xC0B) \
+    f(hpmcounter12   , 0xC0C) f(hpmcounter13   , 0xC0D) f(hpmcounter14   , 0xC0E) f(hpmcounter15   , 0xC0F) \
+    f(hpmcounter16   , 0xC10) f(hpmcounter17   , 0xC11) f(hpmcounter18   , 0xC12) f(hpmcounter19   , 0xC13) \
+    f(hpmcounter20   , 0xC14) f(hpmcounter21   , 0xC15) f(hpmcounter22   , 0xC16) f(hpmcounter23   , 0xC17) \
+    f(hpmcounter24   , 0xC18) f(hpmcounter25   , 0xC19) f(hpmcounter26   , 0xC1A) f(hpmcounter27   , 0xC1B) \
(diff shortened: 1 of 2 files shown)


src/isa/riscv64/local-include/csr.h
```diff
@@ -19,134 +19,305 @@
...
 
+/**
+ * Mapping between CSR name and addr
+ * 
+ * This part follows the order of "CSR Listing" section in Privileged ISA Manual.
+*/
+
+/* Unprivileged CSR */
+/** Unprivileged Floating-Point CSRs **/
+#define CSRS_UNPRIV_FLOAT(f) \
```
poemonsense, line 30: Wrap it with `CONFIG_FPU_NONE`?

poemonsense (approved):
Look good to me after wrapping FP CSRs in CONFIG_FPU_NONE

src/isa/riscv64/local-include/csr.h
cebarobot: Done. Please review it again.

poemonsense (approved):
Just noticed that we are previously not wrapping CSRs under FPU_NONE. I've tested that enabling FPU_NONE in this commit works for running a SPECCPU2006 case gamess_exam29.

poemonsense (commented):
I think I've made a wrong comment to wrap them in FPU_NONE in this PR. I thought they were previously wrapped in FPU_NONE but not in this commit.

But it's good for us to fix this. Thanks

poemonsense:
> > I think I've made a wrong comment to wrap them in FPU_NONE in this PR. I thought they were previously wrapped in FPU_NONE but not in this commit.
> > But it's good for us to fix this. Thanks
> 
> I am considering whether it is better to open a new pr to wrap the fp CSRs. The wrapping commit affects other places, which is not too related to this pr.
> 
> Also, the previous are using a different way to solve the CSRs accessing check (during csr_read() & csr_write()). I think it's better to open another pr and cherry-pick te wrapping commit.

Yes, I agree with you

poemonsense:
I just found that, removing the mtime CSR (actually time) when CONFIG_SHARE is not defined will cause kernel panic when booting a 5.10.167 Linux kernel. @cebarobot Do you know why?

I tested the hello-world-linux in XiangShan, but it works. Do you have some other Linux cases to test it?

time is renamed as mtime to avoid naming conflicts. (Zicntr cannot be actually enabled now)

poemonsense:
> > I just found that, removing the mtime CSR (actually time) when CONFIG_SHARE is not defined will cause kernel panic when booting a 5.10.167 Linux kernel. @cebarobot Do you know why?
> > I tested the hello-world-linux in XiangShan, but it works. Do you have some other Linux cases to test it?
> 
> That's a bit abnormal. XiangShan RTL has not implemented `time` CSR (`RDTIME` pseudoinsturction) yet, but it works well.
> 
> In that case, running `RDTIME` will get trapped into Machine software (like openSBI). Machine software will read the `mtime` MMIO reg of CLINT and emulate the behavior of `RDTIME`. I'm not sure what is going wrong.

I agree with you on the workflow of emulating RDTIME in M-mode software.

I just realized the root cause of this issue. I'm currently using riscv-pk as the bootloader. It has a logical bug in CSR emulation which requires a true in scounteren instead of a false. I'm discussing the issue with the original author.

M-mode software may check the mcounteren/scounteren to ensure whether they should emulate the CSR read. When the counteren bits are set, instructions like rdtime are expected not to raise any illegal instruction exception. Therefore, when counteren is set, M-mode software should not emulate the instruction.

This logic seems to have bugs in riscv-pk.

I checked opensbi. It does not double-check mcounteren or scounteren.

poemonsense:
> > time is renamed as mtime to avoid naming conflicts. (Zicntr cannot be actually enabled now)
> 
> This is really a problem. I'm working on Zicntr and Zihpm extentions and I decided to use `csr_time` instead.
> 
> `mtime` refers to a different MMIO reg, and using `mtime` might cause confusion. Also, `time` is not a machine CSR, so its name should not start with `m`.

Yes. I agree. We may need to come up with another name for this. For example, time_csr?

poemonsense:
> > I agree with you on the workflow of emulating RDTIME in M-mode software.
> > I just realized the root cause of this issue. I'm currently using riscv-pk as the bootloader. It has a logical bug in CSR emulation which requires a true in scounteren instead of a false. I'm discussing the issue with the original author.
> > M-mode software may check the mcounteren/scounteren to ensure whether they should emulate the CSR read. When the counteren bits are set, instructions like rdtime are expected not to raise any illegal instruction exception. Therefore, when counteren is set, M-mode software should not emulate the instruction.
> > This logic seems to have bugs in riscv-pk.
> > I checked opensbi. It does not double-check mcounteren or scounteren.
> 
> This problem is a bit complicated.
> 
> * When a hart does not implement Zicntr/Zihpm, Should the mcounteren/scounteren be read-only zero or WARL?
> * Why openSBI does not check the mcounteren/scounteren? Does it manage the permission inside it self? Or it just let all Supervisor software to access time?
> 
> ISA manual does not discuss futher details. I think we could open an issue about this.

1) mcounteren should be read-only zero. When they can be set to 1, the corresponding read CSR instruction should not cause illegal instructions.

2) This is just a double-check logic by riscv-pk. When the system boots, riscv-pk tries to set all bits of counteren to 1. Therefore, if the CPU does not support any of the performance counters, the corresponding bit will be zero. Then, in the emulation code, riscv-pk double checks the bit. If it is set to 1, the illegal instruction exception should not happen and riscv-pk decides not to emulate it. OpenSBI simply does not double check the counteren and always emulates the instruction.

Even if OpenSBI does have this extra checking logic (given that maybe I was misreading the source code), the system should still run correctly because both XiangShan (https://github.com/OpenXiangShan/XiangShan/commit/e4b1ccac1c8457af674b1afff8d98e13fd5bea53) and NEMU are have a zero write mask for counteren.

Why I encountered the kernel panic: riscv-pk implements this double-check logic incorrectly (since several years ago).

Why previously NEMU, XiangShan, NutShell all work well: previously all of them does not have a write mask for counteren. Thus, at boot when riscv-pk writes ones to counteren, the CSRs are set to all 1.

poemonsense:
One follow-up to fix the write mask of mcounteren: https://github.com/OpenXiangShan/NEMU/pull/346

PR #343, opened 2024-05-17
csr, pmp: allow to implatement 0, 16 or 64 PMP entries

cebarobot:
According to RISC-V priv ISA, implementations may implement zero, 16, or 64 PMP entries; all PMP CSR fields are WARL and may be read-only zero. That means that accessing CSRs of implemented PMP entries should not rasie illegal instruction exception. Also, any CSRs of implemented PMP entries could be read-only zero to indicate that it has no function.

This patch adds the spport of up to 64 PMP entries. Also, this patch allows to choose 0, 16 or 64 of entries are implemented, and allows to choose any number of entries to be active (with function). This patch fixes the wrong behavior of #109, in which version NEMU would raise illegal instruction exception when accessing pmpaddr x (RV_PMP_NUM <= x < 16).

BREAKING: This patch changes some key config of PMP in Kconfig.


poemonsense (approved):

Lemover (approved):
Good Job

PR #346, opened 2024-05-21
priv: fix the write mask of mcounteren

poemonsense:
It should be read-only zero.

### DIFF
--- a/src/isa/riscv64/system/priv.c
+++ b/src/isa/riscv64/system/priv.c
@@ -401,6 +401,8 @@ static inline void csr_write(word_t *dest, word_t src) {
       vsatp->val = MASKED_SATP(src);
   }else if(is_write(hcounteren)){
     hcounteren->val = mask_bitset(hcounteren->val, COUNTEREN_MASK, src);
+  }else if(is_write(mcounteren)){
+    mcounteren->val = mask_bitset(mcounteren->val, COUNTEREN_MASK, src);
   }else if(is_write(scounteren)){
     scounteren->val = mask_bitset(scounteren->val, COUNTEREN_MASK, src);
   }else if (is_write(mstatus)) { mstatus->val = mask_bitset(mstatus->val, MSTATUS_WMASK, src); }


poemonsense:
There are still two issues (one for NEMU, one for software).

1) NEMU incorrectly implements the CSR logic in https://github.com/OpenXiangShan/NEMU/commit/3da78d2cfa4145ff49845dca267a876ce1ee4197

A lot of CSRs are wrapped under RVH, despite they are unrelated to RVH. The code is ugly and difficult to maintain. I believe the current implementation when RVH is disabled is wrong, at least for scounteren/mcounteren. They will not have write masks if RVH is not defined.

2) riscv-pk may behave incorrectly after mcounteren/scounteren are fixed in this PR and previous PRs. I'm reporting the issue. They will be fixed soon in OpenXiangShan and hopefully reported to the upstream quickly

poemonsense:
> 2. riscv-pk may behave incorrectly after mcounteren/scounteren are fixed in this PR and previous PRs. I'm reporting the issue. They will be fixed soon in OpenXiangShan and hopefully reported to the upstream quickly

Fixed in https://github.com/OpenXiangShan/riscv-pk/commit/70c6a5703bee57aea47d58f2a1096b3c368038ad.

I believe the checkpoints based on riscv-pk used by XiangShan should be re-created (using OpenSBI if possible; we are deprecating riscv-pk, though it still works). Also, Linux workloads for NutShell/XiangShan should be rebuilt (using OpenSBI if possible).

cebarobot (commented):
I'm not sure about the order of these three m/h/scointeren CSR. Putting the mcounteren between hcounteren and scounteren is a bit strange...

poemonsense:
> I'm not sure about the order of these three m/h/scointeren CSR. Putting the mcounteren between hcounteren and scounteren is a bit strange...

I agree. These should be fixed along with other CSRs. The H-ext PR introduces too many weird CSR `if` here.

poemonsense:
We can leave this PR not merged. I'm checking with Waterman on the expected behaviors when Zicntr is (not) implemented.

cebarobot (approved):

poemonsense:
I'll leave this unchaged. We need to have writable mcounteren and scounteren, per the requirement of RVA profiles.

Only when mcounteren and scounteren are writable, M-mode software will emulate the counters.

After NEMU CSR is updated next time, we will come back and see what we can do to configure the counteren CSRs.

PR #350, opened 2024-05-24
rv64, csr: add support for Zicntr and Zihpm

cebarobot:
This PR has not been completed yet. This PR should be merged by "rebase".

This PR includes the support of Zicntr and Zihpm:
- [x] Support for CSR `mcycle` and `minstret`.
  * Assuming CPI = 1.
- [x] Support for CSR `cycle`, `time`, `instret`
- [x] Support for Counter-Enable `mcounteren`, `hcounteren`, `scounteren`
- [x] Support for HPM Counter `mhpmcountern` and `hpmcountern`
- [x] Support for Counter-Inhibit `mcountinhibit`


src/isa/riscv64/local-include/csr.h
```diff
@@ -35,9 +35,20 @@
 #endif // CONFIG_FPU_NONE
 
 /** Unprivileged Counter/Timers **/
+#ifdef CONFIG_RV_CSR_time
```
poemonsense, line 38: Why not use upper case letters for macros? CONFIG_RV_CSR_TIME?

src/isa/riscv64/Kconfig
```diff
@@ -48,6 +48,61 @@ config TRIGGER_NUM
   depends on RVSDTRIG
   default 10
 
+menuconfig RV_Zicntr
+  bool "RISC-V Zicntr Extensions for Base Counters and Timers, v2.0"
+  default n
```
poemonsense, line 53: It would be better to default y? `default n` won't boot Linux

poemonsense (approved):
Others look good to me. Please feel free to merge it.

If zicntr has default n, we need to update the default configs. Otherwise, Linux will kernel panic

src/isa/riscv64/Kconfig
cebarobot: Make sense. Also, XiangShan do support `mcycle` and `minstret`.

src/isa/riscv64/local-include/csr.h
cebarobot: Fix.

poemonsense:
Yes, we may need to update Spike as well.

Also, for other macros, why not use upper case letters? It seems strange to use lower case letters in C macros. Users will not read the config using the macro (they are reading the menuconfig), so I think it's ok we use upper case for macros but in menuconfig we use lower case (Zicntr) in the config option.

poemonsense:
> But you are right, using upper case letters in marco const is a convention in C programming language. I could switch to it.

All are good for me. It's up to you to decide it.

Have approved it, and please feel free to merge it anytime. I expect NEMU to not produce kernel panic after this PR

PR #358, opened 2024-06-18
Fix FS,VS and SD fields in mstatus and vsstatus

huxuan0307:
* Set vs and fs of mstatus to OFF(0) when initial hart state.
* Set vstatus.vs dirty when in virtual mode and modifying vector status.
* As specified in RISC-V spec, the mstatus.SD is read-only field and producted from FS, VS or XS is dirty.
* Produce SD field when SD field when copy from REF to DIFFTEST.
* Hardwire SD to 0 when copy from DIFFTEST to REF.

### DIFF
--- a/src/isa/riscv64/system/priv.c
+++ b/src/isa/riscv64/system/priv.c
@@ -758,10 +758,13 @@ static inline void csr_write(word_t *dest, word_t src) {
   }
   else { *dest = src; }
 
+#ifndef CONFIG_FPU_NONE
   if (is_write(fflags) || is_write(frm) || is_write(fcsr)) {
     fp_set_dirty();
     fp_update_rm_cache(fcsr->frm);
   }
+#endif
+
 #ifdef CONFIG_RVV
   if (is_write(vcsr) || is_write(vstart) || is_write(vxsat) || is_write(vxrm)) {
     vp_set_dirty();
--- a/src/isa/riscv64/system/priv.c
+++ b/src/isa/riscv64/system/priv.c
@@ -276,10 +276,9 @@ static inline void update_mstatus_fs() {
 }
 
 inline word_t gen_status_sd(word_t status) {
-  __attribute_maybe_unused__
   mstatus_t xstatus = (mstatus_t)status;
-  bool fs_dirty = MUXNDEF(CONFIG_FPU_NONE, xstatus.fs == EXT_CONTEXT_DIRTY, false);
-  bool vs_dirty = MUXDEF(CONFIG_RVV, xstatus.vs == EXT_CONTEXT_DIRTY, false);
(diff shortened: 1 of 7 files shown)


src/isa/riscv64/system/priv.c
```diff
@@ -267,22 +267,21 @@ word_t inline pmp_tor_mask() {
...
 }
-#ifdef CONFIG_RVH
-static inline void update_vsstatus_sd() {
-  if (hstatus->vsxl == 1)
-    vsstatus->_32.sd = (vsstatus->_32.fs == 3);
-  else
-    vsstatus->_64.sd = (vsstatus->_64.fs == 3);
+
+inline word_t gen_status_sd(word_t status) {
+  __attribute_maybe_unused__
```
cebarobot, line 279: Could this attribute be removed?

pxk27 (approved):

src/isa/riscv64/system/priv.c
```diff
@@ -267,22 +267,21 @@ word_t inline pmp_tor_mask() {
...
+  mstatus_t xstatus = (mstatus_t)status;
+  bool fs_dirty = MUXNDEF(CONFIG_FPU_NONE, xstatus.fs == EXT_CONTEXT_DIRTY, false);
```
cebarobot, line 281: I think there is no need to add a MUX:
* `FS` could be writable even if F-ext/D-ext is not implemented (`CONFIG_FPU_NONE`). NEMU has `CONFIG_RV_MSTATUS_FS_WRITABLE` to control whether `FS` is writeable.
* if `FS` is read-only zero, this would not affect this check.

pxk27 (approved):

cebarobot (changes requested):

src/isa/riscv64/system/priv.c
```diff
@@ -267,22 +267,21 @@ word_t inline pmp_tor_mask() {
...
+  bool vs_dirty = MUXDEF(CONFIG_RVV, xstatus.vs == EXT_CONTEXT_DIRTY, false);
```
cebarobot, line 282: Also this, `VS` is always here. No `CONFIG_RVV` ensures VS is read-only zero.

cebarobot (commented):
Others looks good to me.

src/isa/riscv64/system/priv.c
cebarobot: This is neccessary.

src/isa/riscv64/system/priv.c
huxuan0307: U R right

src/isa/riscv64/system/priv.c
huxuan0307: OK

cebarobot (approved):
LGTM

PR #371, opened 2024-06-27
Fix trigger

huxuan0307:


src/isa/riscv64/system/intr.c
```diff
@@ -170,6 +170,10 @@ word_t raise_intr(word_t NO, vaddr_t epc) {
                     ((v || hld_st_temp) && ((0 <= NO && NO <= 7 && NO != 2) || NO == EX_IPF || NO == EX_LPF || NO == EX_SPF)));
     mstatus->mpv = cpu.v;
     cpu.v = 0;set_sys_state_flag(SYS_STATE_FLUSH_TCACHE);
+#endif
+#ifdef CONFIG_RVSDTRIG
```
cebarobot, line 174: Is this macro correct? It seems that it should be `CONFIG_RV_SDTRIG`.

src/isa/riscv64/system/priv.c
```diff
@@ -97,6 +97,11 @@ static inline bool csr_is_legal(uint32_t addr, bool need_write) {
 #endif
     return false;
   }
+#ifdef CONFIG_RVSDTRIG
```
cebarobot, line 100: Also here.

cebarobot (changes requested):
Please consider to rebase this pr on the latest master.
Others looks good to me.

cebarobot (approved):

PR #372, opened 2024-06-28
rv64v: fix wrong align of get_vreg/set_vreg/get_vlen_max in vld/vst/vldx/vstx.

NewPaulWalker:
The arguments of get_vreg/set_vreg/get_vlen_max should be the coding of emul, rather than coding of lmul in vtype.
For vld and vst, vtype->lmul needs to undergo certain transformations to obtain emul, that is why we should not use vtype->lmul as arguments.
And for vldx and vstx, vtype->lmul also might great than 4, which means that the emul is a fraction and only one vreg is needed. In this time, if use vtype->lmul as arguments, it will cause that using more than 8 (e.g. shifting 1 left by 5 results in 32) vregs in this instruction.


Ziyue-Zhang:
Sorry, this exception check problem has been solved by #374 , we should handle wrong align check before starting vector load or store, because XiangShan check alignment in decode stage. Otherwise, it will cause difftest miss match because nemu may execute some load or store, but XiangShan not execute any load or store yet.

PR #379, opened 2024-07-04
Tmp merge master

sinceforYy:


src/isa/riscv64/system/priv.c
```diff
@@ -541,42 +560,68 @@ static inline void csr_write(word_t *dest, word_t src) {
...
+  else if (is_write(mvien)) { mvien->val = src;}
+  else if (is_write(mvip)) { mvip->val = src;}
+  else if (is_write(hvien)) { hvien->val = src;}
+  else if (is_write(hvictl)) { hvictl->val = src;}
+  else if (is_write(hviprio1)) { hviprio1->val = src;}
+  else if (is_write(hviprio2)) { hviprio2->val = src;}
+  else if (is_write(mtopei)) { mtopei->val = src;}
+  else if (is_write(stopei)) { stopei->val = src;}
+  else if (is_write(vstopei)) { vstopei->val = src;}
+#endif // CONFIG_RV_AIA && CONFIG_RV_IMSIC
```
cebarobot, line 624: It seems that there is no need to write such `else if`s, as there is no side effect of writing to these csrs. There is an `else { *dest = src; }` to handle all normal CSRs.

src/isa/riscv64/system/priv.c
```diff
@@ -541,42 +560,68 @@ static inline void csr_write(word_t *dest, word_t src) {
...
+#if defined CONFIG_RV_AIA && defined CONFIG_RV_IMSIC
+  else if (is_write(miselect)) { miselect->val = src;}
+  else if (is_write(siselect)) { siselect->val = src;}
+  else if (is_write(vsiselect)) { vsiselect->val = src;}
+  else if (is_write(mireg)) { mireg->val = src;}
+  else if (is_write(sireg)) { sireg->val = src;}
+  else if (is_write(vsireg)) { vsireg->val = src;}
+  else if (is_write(mtopi)) { /* do nothing */}
+  else if (is_write(stopi)) { /* do nothing */}
+  else if (is_write(vstopi)) { /* do nothing */}
```
cebarobot, line 614: For read-only CSRs, please just return here.

src/isa/riscv64/system/priv.c
```diff
@@ -834,7 +888,104 @@ word_t csrid_read(uint32_t csrid) {
...
+#endif // CONFIG_RVH
+  }
+#ifdef CONFIG_RVH
+  else if(is_access(hstateen0)) {
+    if((cpu.mode < MODE_M) && (!mstateen0->se0)) { longjmp_exception(EX_II); }
+    else if (cpu.v && mstateen0->se0) { longjmp_exception(EX_VI); }
+  }
+#endif // CONFIG_RVH
+}
+#endif
```
cebarobot, line 909: What does this `#endif` correspond to?

src/isa/riscv64/system/priv.c
```diff
@@ -834,7 +888,104 @@ word_t csrid_read(uint32_t csrid) {
...
+#endif // CONFIG_RV_AIA && CONFIG_RV_IMSIC
+#endif // CONFIG_RVH
+
 static void csrrw(rtlreg_t *dest, const rtlreg_t *src, uint32_t csrid) {
+#ifdef CONFIG_RV_SMSTATEEN
+  smstateen_extension_permit_check(dest, src, csrid);
+#endif // CONFIG_RV_SMSTATEEN
+#ifdef CONFIG_RVH
+#if defined CONFIG_RV_AIA && defined CONFIG_RV_IMSIC
+  aia_extension_permit_check(dest, src, csrid);
```
cebarobot, line 986: Does AIA permision check depend on H-extension?

src/isa/riscv64/system/priv.c
```diff
@@ -834,7 +888,104 @@ word_t csrid_read(uint32_t csrid) {
   return csr_read(csr_decode(csrid));
 }
 
+// VS/VU access statenen should be EX_II when mstateen0->se0 is false.
```
cebarobot, line 891: TYPO: statenen

cebarobot: And I couldn't find the requirement that access to sstateen should raise EX_II. Could you please point out where it is mentioned in the manual?

cebarobot (changes requested):

PR #397, opened 2024-07-15
rv-b: mark B-ext in misa

shinezyy:


poemonsense (approved):

.github/workflows/ci.yml
```diff
@@ -130,7 +130,7 @@ jobs:
           make riscv64-rvh-diff-spike_defconfig
```
cebarobot, line 130: I think this config should be changed to `riscv64-xs-diff-spike_defconfig`. This will solved the current stack overflow problem. But there are still problems because of #407

shinezyy: Yes, but before 407 is merged, I want to add a guard for NEMU as DUT.

shinezyy: > I think this config should be changed to `riscv64-xs-diff-spike_defconfig`. This will solved the current stack overflow problem. But there are still problems because of #407

I enabled RVV in `riscv64-rvh-diff-spike_defconfig`, and it works.
It does not work with `riscv64-xs-diff-spike_defconfig`, here is diff:

```
< # CONFIG_RV_ZICNTR is not set
< # CONFIG_RV_ZIHPM is not set
---
> CONFIG_RV_ZICNTR=y
> # CONFIG_RV_CSR_TIME is not set
> CONFIG_RV_ZIHPM=y
30,31c31,32
< # CONFIG_RV_CSR_MCOUNTINHIBIT_CNTR is not set
< # CONFIG_RV_CSR_MCOUNTINHIBIT_HPM is not set
---
> CONFIG_RV_CSR_MCOUNTINHIBIT_CNTR=y
> CONFIG_RV_CSR_MCOUNTINHIBIT_HPM=y
112c113
< # CONFIG_MEM_COMPRESS is not set
---
> CONFIG_MEM_COMPRESS=y
```

.github/workflows/ci.yml
```diff
@@ -95,6 +95,30 @@ jobs:
...
+          make clean-all
+          make riscv64-xs-diff-spike-agnostic_defconfig
+          make -j
+          set -x
+          ./build/riscv64-nemu-interpreter -b --diff $TEST_HOME/riscv64-spike-GCBVH-ref-so --restore $TEST_HOME/hmmer_retro_6200_0.0378585.gz -r $TEST_HOME/v-gcpt.bin -I 2000000 > crash.log  || exit_code=$?
+          if [ ${exit_code} -eq 0 ]; then echo "Difftest is broken, it should report error!" exit 1; fi
+          match=$(grep "wrong.*=.*ffff" crash.log -c)
+          if [ ${match} -eq 0 ]; then echo "Difftest is broken, it should report at least one agnostic related difference!" exit 1; fi
+          make clean-all
+          set +x
```
shinezyy, line 120: @poemonsense @cebarobot I check the behavior of a failed case here.

PR #408, opened 2024-07-18
rv64f: add difftest for fcsr.

NewPaulWalker:
Since **frm** and **fflags** are not only CSRs but also fields within the **fcsr**, we choose to perform the difftest comparison only on **fcsr** here.


NewPaulWalker:
In addition to support difftest for fcsr, skip difftest for tcontrol CSR in Sdtrig extension when nemu as dut because spike do not implement tcontrol. Then, change H extension ci test config to riscv64-xs-diff-spike_defconfig.

NewPaulWalker:
* Delete difftest skip for tcontrol when nemu as dut because spike-so has implemented tcontrol.
* Trun on Sdtrig extension and set triggers num to 4 to consistent with XiangShan, initialize Sdtrig CSR, fix a tselect write bug.
* Update ci workflow config and diff spike-so.

huxuan0307 (approved):

PR #417, opened 2024-07-24
CI: rearrange NEMU CI

cebarobot:
This PR rearrange NEMU CI:
- ci: all compilation test are in one job

  There are some compilation tests in different jobs, which are moved in compilation test now. This patch also improve the log of compilation test.

- ci: rearrange all ci jobs and add more basic tests

  Previously, the CI tests of NEMU are very simple and strange. There are tests for H and V extensions but isn't any test for basic function.

  This patch rearranges the structure of CI jobs and divides jobs into Basic (stand alone) and Diff with Spike. This patch also adds basic tests such as cputests, riscv-tests, rvv-test and misc-tests. However, some of these tests failed. These failed tests is set to be skipped temporarily.

  Also, NEMU change to use master branch of ready-to-run repo.

- configs: remove useless rvh & rvv configs

  As Vector and Hypervisor extensions were merged into XiangShan master, riscv64-xs_defconfig (and other riscv64-xs defconfigs) has enabled RVH and RVV. Also, the ready-to-run spike-so also supports both RVH and RVV currently. Therefore, riscv64-rvh and riscv64-rvv defconfigs become useless.

  This patch remove all 7 riscv64-rvh & riscv64-rvv defconfigs. Thanks their contribution for developing V and H extensions.

#### WARNING
Some basic tests failed and are set to be skipped temporarily. Please check the skip_tests in ci.yml for details.


poemonsense (approved):

poemonsense:
I think we can default NEMU_HOME to the current path of NEMU. It has been a long time that we have to set it.

Then we don't have to explicitly set it anymore. It seems this env is never set to other values. 

This may be done in another PR.

shinezyy (approved):

PR #442, opened 2024-08-09
add RISC-V Zfh and Zfhmin Extensions V1.0

zmx2018:
rvzfh:add RISC-V Zfh and Zfhmin Extensions V1.0


huxuan0307:
Already merged in #443

PR #450, opened 2024-08-14
rv64: add Zimop extension support

huxuan0307:
* All MOP.R.n and MOP.RR.n only update rd with 0s. This would be changed when any MOP redefined by some other extensions.
* Define all MOP.R.n and MOP.RR.n seperated instruction name for future easier modification, since any one of MOP could be meaningful instruction in the future.
* rtl_zcmop should be defined in local-include/rtl.h instread of include/rtl/pseudo.h
* Since XiangShan has not implemented Zimop extension, disable CONFIG_RV_ZIMOP in config


cebarobot (approved):
LGTM

PR #456, opened 2024-08-17
ci: add difftest with spike for nutshell 

cebarobot:
Currently, NEMU just has difftest CI only on XiangShan defconfigs. This patch add difftest CI on Nutshell, in order to improve NEMU quality in different configs.

This PR also include an compilation fix when PMP is disabled.


poemonsense (approved):
Probably we need to tell users how to build the CI_WORKLOADS somewhere in the doc (or should we simply download the Spike releases?)

PR #473, opened 2024-08-28
build: enable -fPIC to avoid relocation against the same label

Ziyue-Zhang:
* without -fPIC, may have compile error when compile the NEMU as dut

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,7 @@ LDFLAGS += -lz
 endif
 
 ifndef CONFIG_SHARE
+CFLAGS += -fPIC
 LDFLAGS += -lreadline -ldl -pie
 else
 SHARE = 1


poemonsense:
I do not know about `-fPIC`. Any test for this?

Or maybe other people familar with this should have a review. I don't know much about this option

Ziyue-Zhang:
> I do not know about `-fPIC`. Any test for this?
> 
> Or maybe other people familar with this should have a review. I don't know much about this option

when I run make riscv64-xs_defconfig and make, there is a compilation failure. Using -fPIC can solve it.
I also don't know why using -fPIC can solve this problem, fpic is generally used when compiling dynamic link libraries.
Maybe using fpic doesn't really solve the problem.

Folloing is my compile error:
`
+ CXX src/memory/store_queue_wrapper.cpp
mkdir -p resource/softfloat/build
ln -sf /nfs/home/zhangziyue/NEMU/resource/softfloat/repo/build/Linux-x86_64-GCC/softfloat.a resource/softfloat/build/softfloat.a
+ ccache g++ /nfs/home/zhangziyue/NEMU/build/riscv64-nemu-interpreter
/usr/bin/ld: /tmp/ccJtUK8R.ltrans4.ltrans.o: warning: relocation against `.L737' in read-only section `.text'
/usr/bin/ld: /tmp/ccJtUK8R.ltrans4.ltrans.o: in function `tcache_init.constprop.0':
<artificial>:(.text+0x2e57): undefined reference to `.L737'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
collect2: error: ld returned 1 exit status
make: *** [/nfs/home/zhangziyue/NEMU/scripts/build.mk:81: /nfs/home/zhangziyue/NEMU/build/riscv64-nemu-interpreter] Error 1
`

PR #480, opened 2024-08-30
fix(csrrw): refactor csrrw path, execute csr read and write together.

NewPaulWalker:


src/isa/riscv64/local-include/csr.h
```diff
@@ -20,6 +20,13 @@
 #include <common.h>
 #include <memory/vaddr.h>
 
+#define CSRRW_FUNCT3  1
+#define CSRRS_FUNCT3  2
+#define CSRRC_FUNCT3  3
+#define CSRRWI_FUNCT3 5
+#define CSRRSI_FUNCT3 6
+#define CSRRCI_FUNCT3 7
+
```
huxuan0307, line 29: FUNCT3_XXX may be better

huxuan0307 (changes requested):

huxuan0307 (approved):

PR #487, opened 2024-08-31
feat(riscv64): support RISCV smrnmi and ss/smdbltrp extensions

lewislzh:
* support RISCV Resumable Non-Maskable Interrupts extensions
	* add difftest trigger nmi intr support
	* add mnstatus mnepc mnscratch mncause registers
	* add mnret instruction * add nmi intr handler
	* add new config to open this extension
* support RISCV Machine/Superviosr/Virtual Supervisor Level Double Trap extensions
	* add m/henvcfg DTE to control ssdbltrp extension
	* add mstatus.MDT sstatus/vsstatus.SDT to record Double Trap extension and control mie/sie
	* modify m/mn trap handler to support double trap
	* modify mnret, mret, sret to return from double trap
	* add 2 new configs to open these extensions: rely on Smrnmi extension open


src/isa/riscv64/system/priv.c
```diff
@@ -1335,14 +1531,50 @@ static word_t priv_instr(uint32_t op, const rtlreg_t *src) {
...
+      // clear vsstatus.SDT when return to VU
+      vsstatus->sdt = (cpu.mode == MODE_U && cpu.v == 1 ? 0 : vsstatus->sdt);
+      set_sys_state_flag(SYS_STATE_FLUSH_TCACHE);
+#endif // config_RVH
+      mnstatus->mnpp = MODE_U;
+      mnstatus->nmie = 1;
+
+      // clear MDT when mnret to below M
+      mstatus->mdt = (cpu.mode == MODE_M ? mstatus->mdt: 0);
+      mstatus->sdt = (cpu.mode == MODE_M ? mstatus->sdt: 0);
```
huxuan0307, line 1570: mnstatus->mnpp is better than cpu.mode, since cpu.mode is just assigned in this function.
When using state machine design pattern, all state changes should be assigned with current state(mnstatus->mnpp) or the condition of state transfer directly, not the next state of other field(here cpu.mode) which is just assigned to.

huxuan0307 (changes requested):

cebarobot (approved):

huxuan0307 (approved):

PR #489, opened 2024-08-31
fix(cpu-exec): capture instr for mtval in perf_opt

cebarobot:
NEMU updates cpu.instr every instruction only if PERF_OPT is disabled, so that *tval could not get proper instuction info in PERF_OPT mode. This patch update cpu.instr in update_global() function, so that cpu.instr could be udpated properly without performance degradation.


poemonsense (approved):
I don't really know this part of code. But the case described in the commit message looks good to me.

PR #494, opened 2024-09-02
fix(wfi): VU execute wfi should not cause EX_II.

NewPaulWalker:


huxuan0307:
This PR should be tested with the `wfi_tests` in [riscv-hyp-tests](https://github.com/OpenXiangShan/riscv-hyp-tests)

NewPaulWalker:
> This PR should be tested with the `wfi_tests` in [riscv-hyp-tests](https://github.com/OpenXiangShan/riscv-hyp-tests)

This PR is tested OK with the wfi_tests in riscv-hyp-tests.
However, I found that VU mode executing wfi instructions must cause an exception. 
When mstatus->tw is 0, cause virtual instruction exception.
Otherwise, cause the illegal instruction exception.
That is the original code is right!
And the functionality of this pr is equivalent to the functionality of the original code in fact.
So I will close this pr.

PR #498, opened 2024-09-03
feat(pmp): add pmpcfg permit check.

NewPaulWalker:


NewPaulWalker:
This Pr wants to ensure that when accessing odd pmpcfg raise an illegal instruction exception.
However, odd pmpcfg do not exist in rv64 and NEMU can raise an illegal instruction by check exist csr.
So we do not need check odd pmpcfg again.

PR #511, opened 2024-09-06
fix(vmvnr): vmvnr will not set vtype register, no need to check vtype.vill

xyyy1420:


Ziyue-Zhang:
image
image

In spike, vmvnr uses require_vector(true). In require_vector(true), it checks vill.

xyyy1420:
image

emm, but the manual says "without needing to know or change 'vl' or 'vtype'"

Ziyue-Zhang:
https://github.com/riscv-software-src/riscv-isa-sim/issues/1302

Aswaterman thought the spec is wrong

xyyy1420:
but now we already have these instruction which will exec at `vtype.vill == 1`,
maybe we not check `vtype.vill` for now? until both spec and gcc offcially fix this?

PR #514, opened 2024-09-10
fix(ss/smdbltrp) : fix xret and dbltrp handling logic, add reg init

lewislzh:
* correct some code format
	* fix mnstatus mask as not support zicfilp
	* fix m/henvcfg mask according to extension support
	* fix sdt was not correctly assigned when trapping to vs/s mode
	* fix sdt logic of sret

### DIFF
--- a/src/isa/riscv64/system/priv.c
+++ b/src/isa/riscv64/system/priv.c
@@ -214,9 +214,8 @@ static inline word_t* csr_decode(uint32_t addr) {
 #define MSTATUS_WMASK_RVV 0
 #endif
 
-#define MSTATUS_WMASK_SMDBLTRP MUXDEF(CONFIG_RV_SMDBLTRP, (0X1UL << 42), 0)
-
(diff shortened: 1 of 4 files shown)


src/isa/riscv64/system/priv.c
```diff
@@ -1568,6 +1573,7 @@ static word_t priv_instr(uint32_t op, const rtlreg_t *src) {
       if ((cpu.mode == MODE_S && mstatus->tsr) || cpu.mode < MODE_S) {
         longjmp_exception(EX_II);
       }
+      mstatus->sdt = MUXDEF(CONFIG_RV_SSDBLTRP, ((cpu.v == 1) ? mstatus->sdt : 0), 0);
```
huxuan0307, line 1576:
```C++
mstatus->sdt = ISDEF(CONFIG_RV_SSDBLTRP);
#ifdef CONFIG_RV_SSDBLTRP
if (mstatus->spp = MODE_U && mstatus->spv = 1)
  vsstatus->sdt = mstatus->spp = MODE_U && mstatus->spv = 0;
#endif
```

huxuan0307 (changes requested):

huxuan0307 (approved):

PR #531, opened 2024-09-19
fix(snprintf),refactor(small/fast log): fix truncated writing warning, refactor output to log file

xyyy1420:
When snprintf dest buffer is a static array with fixed length, and src
also a static array with fixed length, compiler will tell if there will
be a write truncation and warn about it. This commit could fix this warn.

The logic for outputting the log to a file was a bit hard to understand,
refactored this part of the code.

### DIFF
--- a/include/cpu/decode.h
+++ b/include/cpu/decode.h
@@ -44,6 +44,10 @@ enum {
   INSTR_TYPE_I, // indirect
 };
 
+IFDEF(CONFIG_DEBUG, extern char log_bytebuf[80];)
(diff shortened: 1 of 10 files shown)


shinezyy:
LGTM

For others to understand this patch, can you attach a log or screenshot of the error before this patch?

xyyy1420:
iwEcAqNwbmcDAQTRC80F0QVvBrBE5IIaANmh4QbSYe0JOV4AB9J3OnDXCAAJomltCgAL0gAENQU png_720x720q90

before this patch, when i use `riscv64-xs-ref_defconfig` and enable `debug feature`, `trace instructions`, `rich log` config, i meet this error

shinezyy (approved):

PR #532, opened 2024-09-19
feat(svpbmt): support PBMTE & bump ready-to-run

Maxpicca-Li:
1. set the default PBMTE 1 when PBMT is supported and raise page fault when PBMTE is 0 and PTE.PBMT is not 0.
2. update `csr_read` and `csr_write` to resolve the case when henvcfg.pbmt is read-only 0 and ignored write.
3. fix  a logic error during read-only processing.
4. bump ready-to-run


src/isa/riscv64/init.c
```diff
@@ -91,6 +91,11 @@ void init_isa() {
 #endif
 #endif
 
+#ifdef CONFIG_RV_SVPBMT
+  menvcfg->pbmte = 0;
+  henvcfg->pbmte = 0;
```
cebarobot, line 96: Should `pbmte` be initalized to 0? I remember it should be 1?

Maxpicca-Li: By doing a diff with spike, it is found that the default value needs to be 0.

lewislzh (approved):

cebarobot (approved):

huxuan0307 (approved):

PR #537, opened 2024-09-21
fix(csr): check inst exception for Zicbom & Zicboz

cailuoshan:


good-circle:
@Tang-Haojin May this PR should be merged into master before release?

src/isa/riscv64/instr/rvcbo/exec.h
```diff
@@ -23,6 +23,13 @@
 #define CACHE_BLOCK_OPS          (CACHE_BLOCK_SIZE / CACHE_OP_SPLIT_SIZE)
 
 def_EHelper(cbo_zero) {
+  // check illegal instruction exception
+  if(!cpu.v && ((cpu.mode != MODE_M && !menvcfg->cbze) || (cpu.mode == MODE_U && !senvcfg->cbze))){
```
sinceforYy, line 27: cbo-cbze

The CBO Spec only says it is not in M ​​mode, maybe it is in HS/HU/VS/VU mode.

cailuoshan: Make sense, already fix it.

sinceforYy (approved):

PR #538, opened 2024-09-22
feat(force_raise_pf): dut can guide the REF's {x}tval

Anzooooo:
Due to the complexity of vector access, it is difficult to judge the situation of triggering pf. So given that the DUT is already used to guide REF in setting other CSR values, here the setting of csr when force_raise_pf is guided by the DUT.


pxk27 (approved):

cebarobot (approved):
LGTM. But do we have any ci for vector *tval?

Anzooooo:
> LGTM. But do we have any ci for vector *tval?

Not yet.  ci will be added after the vector exception is complete.

PR #545, opened 2024-09-24
fix(csr, RVH): modify hstatus.vsbe to RO to be same with XiangShan

pxk27:


sinceforYy (approved):

sinceforYy:
For code readability and maintainability, it may be better to use `define` for constants.

sinceforYy (approved):

PR #548, opened 2024-09-24
fix: fix some potentially problematic code

xyyy1420:
- fix initialization of union vars, using  declaration, assignments to define vars, not forced type conversion
- fix initialization of 'bool is_write' crossing 'goto bad', define 'bool is_write' at first
- fix passing to parameter to function, the 'update_global' func take no arguments
- fix some vars type mismatch


shinezyy:
This patch is preparing for re-add the support of C++

shinezyy (approved):
LGTM

PR #551, opened 2024-09-25
fix(diff): use `nemu_large_memcpy` only when init

happy-lx:
Add a `difftest_memcpy_init` function. When NEMU is used as a REF for difftest, during the initialization phase, difftest uses `difftest_memcpy_init` to initialize the memory of NEMU, and during the execution phase, `difftest_memcpy` is used for data transfer between NEMU and difftest.

This is because the `difftest_memcpy_init` function uses `nemu_large_memcpy`. This function can effectively reduce physical memory consumption, but it can only guarantee correctness when dest is all 0, so we only use it in the memory initialization phase now. Currently, `difftest_memcpy` uses memcpy to transfer data, which may require more physical memory during simulation, but the correctness can be guaranteed.

### DIFF
--- a/src/cpu/difftest/ref.c
+++ b/src/cpu/difftest/ref.c
@@ -26,7 +26,7 @@ extern void load_flash_contents(const char *flash_img);
 
 #ifdef CONFIG_LARGE_COPY
 #ifndef CONFIG_USE_SPARSEMM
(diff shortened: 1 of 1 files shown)


poemonsense (approved):
It requires further code style optimization. But it looks good to me

poemonsense (approved):
LGTM

PR #558, opened 2024-09-29
fix(paddr): check mmio misalign based on paddr

cebarobot:
Currently, mmio misaligned memory access is checked in paddr read/write process. In this process, the check should based on physical address, because the paddr and vaddr may not match of some requests, such as page table walk. This patch fixes it.

### DIFF
--- a/src/memory/paddr.c
+++ b/src/memory/paddr.c
@@ -147,9 +147,9 @@ static inline void raise_read_access_fault(int type, vaddr_t vaddr) {
   raise_access_fault(cause, vaddr);
 }
 
-static inline void isa_mmio_misalign_data_addr_check(vaddr_t vaddr, int len, int type, int is_cross_page) {
-  if (unlikely((vaddr & (len - 1)) != 0) || is_cross_page) {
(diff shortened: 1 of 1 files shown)


Ziyue-Zhang:
Pgae offset of vaddr and paddr are same, which is addr & PAGE_SIZE. Why we need to distinguish them?

Ziyue-Zhang:
> Actually, PTW never generates misaligned memory access.
> 
> In before design, misaligned check are all done in MMU check, which is before the PTW process. The physical memory access called by PTW would not be checked. All checked memory accesses are original memory access from instruction fetching, loading or storing.
> 
> In current design, MMIO misaligined are located in `paddr_read()` and `paddr_write()`, where the memory accesses could either from IF/LD/ST or from PTW. For accesses from PTW, their vaddr and paddr are not "matched".

When page table walk, the physical address is in physical memory range, the `paddr_read()` or 'paddr_write()` doesn't call `isa_mmio_misalign_data_addr_check()`

Ziyue-Zhang (approved):

PR #573, opened 2024-10-10
refactor(csrrw): rewrite csr read/write using switch-case instead of multiple if-else.

NewPaulWalker:
The previous `csr_read` and `csr_write` functions contained a lot of if-else-if structures, making maintenance difficult.
I have rewritten these two functions using switch-case statements, including:
* For specific CSR, all related operations are placed under the corresponding case, while the typical CSRs are handled using the default case.
* For a series of consecutive CSRs, like pmpaddr, using switch-case was less convenient, so I kept the original if checks.
* Additionally, I removed some always-true code, dead code and the read-write code for some typical CSRs that were occupying a separate case.

### DIFF
--- a/src/isa/riscv64/difftest/dut.c
+++ b/src/isa/riscv64/difftest/dut.c
@@ -23,9 +23,6 @@
 // csr_prepare() & csr_writeback() are used to maintain 
 // a compact mirror of critical CSRs
 // For processor difftest only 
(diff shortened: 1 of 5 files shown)


NewPaulWalker:
Using case range to handle a contiguous address range of CSR instead of if.

huxuan0307 (approved):

PR #581, opened 2024-10-13
fix(priv): prevent `mstatus.MPP`changing to 2

Tang-Haojin:
If an instruction tries to change `mstatus.MPP` to 2, ignore the writing of this field.

### DIFF
--- a/src/isa/riscv64/system/priv.c
+++ b/src/isa/riscv64/system/priv.c
@@ -192,6 +192,10 @@ static inline word_t* csr_decode(uint32_t addr) {
 // base mstatus wmask
 #define MSTATUS_WMASK_BASE (0x7e19aaUL)
 
+// mstatus.mpp mask
+#define MSTATUS_OFFSET_MPP 11
+#define MSTATUS_MASK_MPP (0x3UL << MSTATUS_OFFSET_MPP)
+
 // FS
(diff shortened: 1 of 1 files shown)


src/isa/riscv64/system/priv.c
```diff
@@ -1326,6 +1330,10 @@ static inline void csr_write(word_t *dest, word_t src) {
       }
     }
   #endif //CONFIG_RV_SSDBLTRP
+    // mstatus.MPP cannot hold MODE_HS=2
```
huxuan0307, line 1333: There is nothing named MODE_HS. 2 is just a reserved code for MPP

src/isa/riscv64/system/priv.c
```diff
@@ -1326,6 +1330,10 @@ static inline void csr_write(word_t *dest, word_t src) {
...
+    if ((src & MSTATUS_MASK_MPP) == (MODE_HS << MSTATUS_OFFSET_MPP)) {
+      mstatus_wmask &= ~MSTATUS_MASK_MPP;
+    }
```
huxuan0307, line 1336: Use `((mstatus_t)src).mpp == 0b10` instead

src/isa/riscv64/system/priv.c
Tang-Haojin: Right. This is an internal name in NEMU. It is used everywhere.

huxuan0307 (approved):

PR #600, opened 2024-10-24
NEMU:support direct reading of *.elf workload

zhaoxuying:
open CONFIG_USE_SPARSEMM macro definition


shinezyy:
Seems strange.
1. Please give enough motivation on enabling SPASE MEM by default
2. Why fno-strict-aliasing is set?

poemonsense (changes requested):
1) The implementation of sparse mem is dirty. Supporting elf is easy but the original code owner was choosing a wrong way to implement it. We must refactor it before officially enabling the ELF support.

2) I don't see any necessity to enable sparse mem in the REF mode. In the REF mode, memory contents are copied manually from the DUT, where NEMU does nothing but `memcpy`.

zhaoxuying:
close this PR, use another method to read elf workload.

PR #608, opened 2024-10-28
refactor: complete migration of isa-related logic to isa.mk

xyyy1420:


shinezyy (commented):
LGTM. But can you add some motivation why we do this? just for code clean-up or for other features?

xyyy1420:
> LGTM. But can you add some motivation why we do this? just for code clean-up or for other features?

The original build command is `make ISA=xxx`,  `ISA` variable is passed directly into make via the command, but now we're using menuconfig, ISA variable is passed into make by writting auto.conf file, so in order for tools/xxx/Makefile to work, we need to organize the logic related to the ISA variables into isa.mk and then include it in tools/xxx/Makefile

shinezyy (approved):
plz squash commits, and add motivation into commit message

PR #613, opened 2024-10-29
fix(build): extract .a files before running ar

poemonsense:
.a libraries cannot be recursively loaded. We extract the .o files before linking them together.

This fixes the staticlib build errors when softfloat is enabled.

### DIFF
--- a/scripts/build.mk
+++ b/scripts/build.mk
@@ -82,9 +82,16 @@ $(BINARY): $(OBJS) $(LIBS)
 
 staticlib: $(BUILD_DIR)/lib$(NAME).a
 
+TEMP_EXTRACT_DIR = $(BUILD_DIR)/build-lib$(NAME)-tmp
 $(BUILD_DIR)/lib$(NAME).a: $(OBJS) $(LIBS)
+	@mkdir -p $(TEMP_EXTRACT_DIR)
+	@echo Extracting object files from libraries...
+	@for lib in $(LIBS); do \
+	    ar x $$lib --output=$(TEMP_EXTRACT_DIR); \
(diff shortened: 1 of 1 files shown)


scripts/build.mk
```diff
@@ -82,9 +82,16 @@ $(BINARY): $(OBJS) $(LIBS)
...
 
+TEMP_EXTRACT_DIR = $(BUILD_DIR)/build-lib$(NAME)-tmp
 $(BUILD_DIR)/lib$(NAME).a: $(OBJS) $(LIBS)
+	@mkdir -p $(TEMP_EXTRACT_DIR)
+	@echo Extracting object files from libraries...
+	@for lib in $(LIBS); do \
+	    ar x $$lib --output=$(TEMP_EXTRACT_DIR); \
+	done
+	@ar rcs $@ $(TEMP_EXTRACT_DIR)/*.o $(OBJS)
 	@echo + AR $@
```
cebarobot, line 93: Maybe `echo` should before `ar rcs`?

By the way, why need to remove extract dir?

cebarobot (approved):

scripts/build.mk
poemonsense: > Maybe `echo` should before `ar rcs`?

Yes. Fixed.

> By the way, why need to remove extract dir?

The extracted .o files here are useless. The temp directory here is only for the linking. Original .o files exist in the build directory (created during building the softfloat library). We extract it again here only to simplify the dependencies.

cebarobot: Ok, get it. LGTM

PR #632, opened 2024-11-04
fix(dbltrp): pass critical-error to difftest, when NEMU acts as REF

lewislzh:


src/isa/riscv64/system/intr.c
```diff
@@ -108,7 +108,10 @@ word_t raise_intr(word_t NO, vaddr_t epc) {
 #endif
 #ifdef CONFIG_RV_SMRNMI
 #ifdef CONFIG_SHARE
-  if (!mnstatus->nmie){return 0; } // this will assert in difftest
+  if (!mnstatus->nmie){
+    IFDEF(CONFIG_RV_SMDBLTRP,cpu.critical_error = true);
+    return 0;
+  } // this will compare in difftest
 #else
   Assert(mnstatus->nmie, "critical error: trap when nmie close");
```
good-circle, line 116: May should not use assert here? If that's what the software is designed to do, we should print some error message for software and hit good trap?

lewislzh: got it , i have fixed it

good-circle (approved):

PR #638, opened 2024-11-06
fix(trigger): extern label should not be inside switch case block.

Donkey-Hao:
`extern Decode *prev_s;` inside a switch case block, which is not allowed because it is a declaration and not an executable statement.

### DIFF
--- a/src/isa/riscv64/system/trigger.c
+++ b/src/isa/riscv64/system/trigger.c
@@ -315,7 +315,6 @@ bool trigger_reentrancy_check() {
 }
 
 void trigger_handler(const trig_type_t type, const trig_action_t action, word_t tval) {
-  extern Decode *prev_s;
   switch (action) {
     case TRIG_ACTION_NONE: /* no trigger hit, do nothing */; break;
     case TRIG_ACTION_BKPT_EXCPT:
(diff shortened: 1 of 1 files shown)


Donkey-Hao:
When I make riscv64-xs-ref_defconfig, error occurred:
```
+ CC src/isa/riscv64/instr/rvv/vcompute_impl.c
+ CC src/isa/riscv64/instr/rvv/vreg_impl.c
src/isa/riscv64/system/trigger.c: In function ‘trigger_handler’:
src/isa/riscv64/system/trigger.c:328:11: error: a label can only be part of a statement and a declaration is not a statement
  328 |           extern Decode *prev_s;
      |           ^~~~~~
make: *** [/home/xs-env/NEMU/scripts/build.mk:62: /home/xs-env/NEMU/build/obj-riscv64-nemu-interpreter-so/src/isa/riscv64/system/trigger.o] Error 1
make: *** Waiting for unfinished jobs....
```
cc = gcc version 9.5.0 (Ubuntu 9.5.0-1ubuntu1~22.04)
commit id: fce68dcb

NewPaulWalker:
In our work environment, the GCC version is 11.4.0, and this version allows a declaration to follow a label, which is not allowed in GCC 9.5.0.
Although this PR is okay, I suggest that you update and use GCC 11.4.0.
Additionally, there is another feasible fix for GCC 9.5.0, that is enclose the declaration immediately following the case label with {}.
```c
case TRIG_TYPE_ICOUNT:
{
  extern Decode *prev_s;
  prev_s->pc = cpu.pc;
}
```

NewPaulWalker (approved):

Donkey-Hao:
Thanks for your suggestion, bug the code should be robust and not rely solely on a high-version GCC compiler.
```
case TRIG_TYPE_ICOUNT:
{
  extern Decode *prev_s;
  prev_s->pc = cpu.pc;
}
```
Yes, this is elegant.

PR #657, opened 2024-11-12
fix(exception): fix exception type raised in paddr_write.

NewPaulWalker:
NEMU should call raise_access_fault directly in paddr_write, rather than raise_read_access_fault.


Anzooooo (approved):

ha0lyu:
Hi @NewPaulWalker,
Please link the issue when you open PRs to fix bugs, which will be helpful for people to check issues.
Thanks.

PR #659, opened 2024-11-14
chore: add new script to convert the built fw_payload.bin into a binary capable of generating checkpoint

xyyy1420:
With this script, it is easy to convert any built fw_payload.bin into a binary file capable of generating checkpoints, which can then be combined with the existing scripts in `scripts/checkpoint_example` to generate checkpoints more easily

### DIFF
--- a/scripts/checkpoint_example/rebuild_fw_payload.sh
+++ b/scripts/checkpoint_example/rebuild_fw_payload.sh
@@ -31,6 +31,7 @@ export RESULT=extract_fw_payload
 
 mkdir $RESULT
 
+# note: default payload offset in opensbi is 2M, if you have modified FW_PAYLOAD_OFFSET, modify the next line of command to match this value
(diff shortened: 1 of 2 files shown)


scripts/checkpoint_example/rebuild_fw_payload.sh
```diff
@@ -0,0 +1,51 @@
...
+  fi
+done < "$REPO_LIST"
+
+cd nemu_board/dts && bash build_single_core_for_nemu.sh && cd $WORKSPACE
+
+export RESULT=extract_fw_payload
+
+mkdir $RESULT
+
+dd if=$FW_PAYLOAD of=$RESULT/kernel.Image bs=1M skip=2
```
shinezyy, line 34: Add some notes when they are not linking Linux at 2MB

shinezyy (approved):

PR #694, opened 2024-12-02
fix(checkpoint scripts): use zstd format by default

xyyy1420:


shinezyy:
grammar issue: use zstd format by default

xyyy1420:
> grammar issue: use zstd format by default

already fixed

PR #718, opened 2024-12-13
fix(is_ctrl): set is_ctrl is true when get rtl_priv_jr exec

xyyy1420:


shinezyy (approved):

shinezyy:
Maybe, you can add some addtional info in the body of commit message:

fix(is_ctrl): set is_ctrl is true when get rtl_priv_jr exec

This patch mark mret as control instruction

xyyy1420:
> Maybe, you can add some addtional info in the body of commit message:
> 
> fix(is_ctrl): set is_ctrl is true when get rtl_priv_jr exec
> 
> This patch mark mret as control instruction

already fixed

PR #724, opened 2024-12-15
feat(profiling): record instr count as base

cebarobot:
Previously, the global instruction counter variable (g_nr_guest_instr) would be reset to 1 when executing nemu_trap 0x101 (start profiling) or receiving ctrl-c signal to start checkpointing, so that profiling and profiling mechanism could create checkpoints in specified intervals.

However, g_nr_guest_instr is a very important variable, which are and will be used in CSRs (minstret, mtime, mcycle) and timer device (clint). It's not good to reset it during running.

This patch modifies the original reset_inst_counters(): it now records current instr count as base with get_abs_instr_count(), and rename the function to start_profiling(). per_bb_profile() is also modified to use the difference between origin count and base as new instr count.

### DIFF
--- a/include/profiling/profiling_control.h
+++ b/include/profiling/profiling_control.h
@@ -21,6 +21,7 @@ extern int checkpoint_state;
 extern bool checkpoint_restoring;
 extern uint64_t checkpoint_interval;
 extern uint64_t warmup_interval;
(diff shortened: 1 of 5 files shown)


xyyy1420 (approved):
LGTM

PR #727, opened 2024-12-18
feat: more accurate instruction counting

cebarobot:
This patch refactors the instruction counting logic in cpu-exec.c to improve its accuracy. It serves as the foundation for implementing the CSR mcycle/cycle, mtime/time and minstret/instret, and is also essential for supporting deterministic time interrupt.

This patch, based on Zihao Yu's talk at RVSC 2021, categorizes instruction counting into three level: DISABLED, BY_BB, BY_INSTR. Corresponding adjustments are made to the Kconfig options.

This patch improves the BY_BB and BY_INSTR instruction counting logic and fixes several issues:

* The variable n and the return value of execute() are deprecated, with the global variable n_remain fully replacing both. Since NEMU could longjmp out of execute(), so return value is no longer meaningful.

* The n_batch variable has been made global, eliminating the need to compute it in get_abs_instr_count() and update_instr_cnt(). Also, n_remain is initialized at the start of execute(), and both n_batch and n_remain are cleared in update_instr_cnt() to avoid duplicate updating.

* The update_instr_cnt() function now handles all n_remain_total updates based on the actual number of instructions executed by execute(), both in PERF_OPT and !PERF_OPT modes. Redundant calls to update_instr_cnt() are removed.

* For BY_BB, rtl_priv_next caused a situation where BB has not ended but execute() is about to end, leading to loss instruction counts. This is now fixed by settling instruction counting immediately before the tcache flush and recording unsettled instructions for upcoming tcache flushing (in case of interrupts).

* For BY_INSTR && PERF_OPT, goto end_of_loop skipped per-inst action part, resulting in missing count for that instruction. It is fixed.

* FOR BY_INSTR && !PERF_OPT, interrupt and state checks may break out of the main loop, causing a miss decrement of n_remain, which led to single step issue. It is fixed.

This patch also formats some codes and adds log output for debugging.

BREAKING: This patch modifies the Kconfig options for instruction counting. All default configs have been updated.


Tang-Haojin (approved):
LGTM. Too Strong!

PR #730, opened 2024-12-18
fix(rtl): incorrect usage of rtl_jr in other rtl

cebarobot:
rtl_j, rtl_jr and rtl_jrelop are special rtl that could not only change guest's (workload's) execution flow but also affect host's (NEMU's) execution flow. They are defined seperately in PERF_OPT mode (as marcos in cpu-exec.c) and normal mode (as functions in rtl-basic.h). They are restricted to being used only directly within EHelper.

rtl_trap contained some legacy code where `rtl_jr` was used incorrectly. This patch removes that improper usages of rtl_jr. Each call to rtl_trap are reviewed to make sure its functionality remains unaffected. Now, every EHelper that calls rtl_trap also calls rtl_priv_jr afterward. (Although mips32 has already been broken and abandoned, a rtl_priv_jr is added to EHelper of syscall to keep consistency)


xyyy1420 (approved):
LGTM

PR #731, opened 2024-12-18
fix(profiling): subtract base from instr count

cebarobot:
PR #724 introduced an intrcution counting base for profiling, but at the time, the simpoint_profiling() in normal rtl_j and rtl_jr was not adjusted. This patch addresses that oversight.


xyyy1420 (approved):
LGTM

PR #734, opened 2024-12-20
feat: deterministic risc-v timer interrupt

cebarobot:
Previously, NEMU relied on the host alarm mechanism (using Linux's setitimer) to update the RTC and trigger timer interrupts, with host- time-based RTC. This approach introduced randomness in each run, leading to differences in traces and results. Although NEMU provided a DETERMINISTIC option, it simply incremented the RTC by a fixed value whenever the CLINT register was read or written, which could not function properly.

This patch refactors the RISC-V timer system and the DETERMINISTIC option. The RISC-V timer and CLINT have now been decoupled: the RISC-V timer is responsible for maintaining the RTC clock (mtime) and handling timer interrupts, while the CLINT manages the MMIO interface for the M-mode timer. When DETERMINISTIC is enabled, the mtime is updated based on instruction counts within the cpu_exec() or execute() loop. Each time NEMU checks for interrupts, it generates timer interrupts based on the current mtime. Furthermore, the patch disables the alarm mechanism entirely when DETERMINISTIC is enabled.

This patch enhances support for the WFI instruction. When executing WFI, NEMU advances the mtime to the next timer interrupt by calculating the remaining ticks and adjusting accordingly. For workloads like KVM, this reduces runtime by half and instruction count by 70%.

This patch also fixes a bug: when v=1, accessing CSR time should return the value of mtime + htimedelta. This ensures that virtualization workloads, such as KVM, now behave as expected.

BREAKING: This patch modifies the Kconfig options for timer and deterministic. Some key default configs have been updated.


Tang-Haojin (approved):

sinceforYy (approved):
LGTM.

eastonman:
I am a bit concerned about 128 instructions per mtime cycle. If I understand it correctly, for a 3 GIPS core, mtime frequency is like 24MHz. This will require device tree setting correct mtimer frequency (at least similar frequency). If the timer frequency used by the kernel is way off (e.g. 1GHz), some kernel operation may take excessive time (e.g. unaligned benchmark).

Maybe some suggestions on software configuration here or left some discussion here.

Kconfig
```diff
@@ -359,6 +355,23 @@ config TIMER_CLOCK_GETTIME
...
 
+config DETERMINISTIC
+  bool "Make the behavior of NEMU deterministic"
+  default y
+  help
+    If enabled, NEMU's execution will be fully deterministic, ensuring that
+    running same workload multiple times produces identical instruction trace
+    and results.
+
+    If enabled, the udpate of timer and devices will no longer depend on host
```
eastonman, line 366: Typo

Tang-Haojin:
> > I am a bit concerned about 128 instructions per mtime cycle. If I understand it correctly, for a 3 GIPS core, mtime frequency is like 24MHz. This will require device tree setting correct mtimer frequency (at least similar frequency). If the timer frequency used by the kernel is way off (e.g. 1GHz), some kernel operation may take excessive time (e.g. unaligned benchmark).
> > 
> > Maybe some suggestions on software configuration here or left some discussion here.
> 
> I chose 128 as default value because, in XiangShan simulation environment, RTC is set to 1/100 of the main frequency. Moreover, most linux workloads used by XiangShan design team use a timebase-frequency of 30 MHz.
> 
> I understand your concern. If I recall correctly, there is a spec which asks mtime tick to be 1 GHz. However, since a config option is provided, I think this won't be a big issue.

The Server Spec requires the unit of mtime is 1ns, but not requires the frequency of a single tick is 1ns. It may also be 30ns or something else.

eastonman:
Looks good to me. Will try this patch on DeterLoad to see if it solves the problem mentioned before.

Kconfig
cebarobot: Fixed.

eastonman (approved):
Ok, LGTM

xyyy1420 (approved):

PR #737, opened 2024-12-24
configs(cpt): disable debug spec on cpt cfg

eastonman:
mcontrol6_trigger_check is consuming about 30% of the runtime, I think currently we don't need debug features on checkpoints.


cebarobot (approved):
LGTM. But please let backend team to review this too.

eastonman:
Debug CSR has to remain, I have made a change that only disables sdtrig.

cebarobot (approved):

Tang-Haojin:
Generally LGTM. We may figure out why this decrease the performance greatly later. @wissygh 

Have you tried the generated checkpoints? Can these be used on XS successfully? @eastonman

wissygh:
> Generally LGTM. We may figure out why this decrease the performance greatly later. @wissygh
> 
> Have you tried the generated checkpoints? Can these be used on XS successfully? @eastonman

Trigger may cause breakpoint exception in nonDmode.

eastonman:
mcontrol6 trigger checks every possible match, like PC, load addr or even load value. I think the implementation is not ideal, maybe skip matching earlier when the triggers are not enabled? @wissygh 
This passes the zstd ckpt generate and restore CI. I think if the checkpoint can be restored on NEMU, it can be restored on XiangShan. @Tang-Haojin 
I am now generating a suite of checkpoints using DeterLoad. I will report if any error occurs.

eastonman:
I have tested one ckpt on XiangShan with difftest, and it seems no problem.

PR #714, opened 2024-12-10
support generate checkpoint to flash file and memory file, and support restore from these files

xyyy1420:
* Submodule
   * add protobuf
* take cpt to flash
  * support for generating checkpoints to flash, thus avoiding the need to link payloads with libcheckpointAlpha.
  * two files will be generated, a flash image and a memory image
  * support difftest from flash
  * change the restore option to specify a flash image, since both checkpoint images and binary are essentially memory mirrors
* difftest
  * support override ref model pmp and pmpcfg regs
  * support override ref model flash contents
  * tried to fix difftest attach but failed, because of spike has many non-register states which are hard to fix
* CI
  * because protobuf is added, submodule is initialized by default in CI
  * add take cpt with flash and restore cpt with flash
  * migrating nutshell-spike-ref from local to ready-to-run
* config file
  * fix the address setting in NEMU riscv64-nutshell-diff-spike_defconfig according to the flash address set in spike.
  * add two configuration files for generating checkpoints to be placed in flash
* MMIO
  * add three flags to indicate which operations need to skip difftest ref model

PR #740, opened 2024-12-26
feat(device): Supports selective difftest of mmio address space

xyyy1420:
split from #714

### DIFF
--- a/include/device/map.h
+++ b/include/device/map.h
@@ -53,7 +53,7 @@ static inline int find_mapid_by_addr(IOMap *maps, int size, paddr_t addr) {
 }
 
 void add_pio_map(const char *name, ioaddr_t addr,
-        void *space, uint32_t len,  io_callback_t callback);
+        void *space, uint32_t len, io_callback_t callback);
 void add_pio_map_with_diff(const char *name, ioaddr_t addr,
         void *space, uint32_t len, int mmio_diff_type, io_callback_t callback);
 
--- a/include/device/map.h
+++ b/include/device/map.h
@@ -34,7 +34,7 @@ typedef struct {
   paddr_t low;
   paddr_t high;
   void *space;
-  int mmio_diff_flag;
+  int mmio_diff_type;
   io_callback_t callback;
 } IOMap;
 
@@ -53,9 +53,9 @@ static inline int find_mapid_by_addr(IOMap *maps, int size, paddr_t addr) {
 }
 
 void add_pio_map(const char *name, ioaddr_t addr,
-        void *space, uint32_t len, int mmio_diff_flag, io_callback_t callback);
+        void *space, uint32_t len, int mmio_diff_type, io_callback_t callback);
 void add_mmio_map(const char *name, paddr_t addr,
-        void *space, uint32_t len, int mmio_diff_flag, io_callback_t callback);
+        void *space, uint32_t len, int mmio_diff_type, io_callback_t callback);
 
 word_t map_read(paddr_t addr, int len, IOMap *map);
 void map_write(paddr_t addr, int len, word_t data, IOMap *map);
--- a/include/device/map.h
+++ b/include/device/map.h
@@ -23,7 +23,7 @@ typedef enum {
   MMIO_READ = 1,
   MMIO_WRITE = 2,
   MMIO_EXEC = 4
-} SKIP_REF_FLAG;
+} MMIO_DIFF_FLAG;
 
 typedef void(*io_callback_t)(uint32_t, int, bool);
 uint8_t* new_space(int size);
@@ -34,7 +34,7 @@ typedef struct {
   paddr_t low;
   paddr_t high;
   void *space;
-  int skip_ref_flag;
+  int mmio_diff_flag;
   io_callback_t callback;
 } IOMap;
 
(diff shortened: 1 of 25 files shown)


xyyy1420:
WIP

include/device/map.h
```diff
@@ -18,6 +18,13 @@
 
 #include <cpu/difftest.h>
 
+typedef enum {
+  SKIP_FREE = 0,
+  MMIO_READ = 1,
+  MMIO_WRITE = 2,
+  MMIO_EXEC = 4
+} SKIP_REF_FLAG;
```
shinezyy, line 26: Is it approriate to call this enum "SKIP_REF_FLAG"?

I think it's mmio_diff_type?

src/device/io/map.c
```diff
@@ -46,6 +46,9 @@ static inline void invoke_callback(io_callback_t c, paddr_t offset, int len, boo
 word_t map_read(paddr_t addr, int len, IOMap *map) {
   assert(len >= 1 && len <= 8);
   check_bound(map, addr);
+  if((MMIO_READ & map->skip_ref_flag) != 0) {
```
shinezyy, line 49: replace 0 with skip_ref

src/device/io/map.c
```diff
@@ -54,6 +57,9 @@ word_t map_read(paddr_t addr, int len, IOMap *map) {
 void map_write(paddr_t addr, int len, word_t data, IOMap *map) {
   assert(len >= 1 && len <= 8);
   check_bound(map, addr);
+  if((MMIO_WRITE & map->skip_ref_flag) != 0) {
```
shinezyy, line 60: replace 0 with skip_ref ?

src/device/io/port-io.c
```diff
@@ -22,11 +22,11 @@ static IOMap maps[NR_MAP] = {};
 static int nr_map = 0;
 
 /* device interface */
-void add_pio_map(const char *name, ioaddr_t addr, void *space, uint32_t len, io_callback_t callback) {
+void add_pio_map(const char *name, ioaddr_t addr, void *space, uint32_t len, int skip_ref_flag, io_callback_t callback) {
   assert(nr_map < NR_MAP);
   assert(addr + len <= PORT_IO_SPACE_MAX);
   maps[nr_map] = (IOMap){ .name = name, .low = addr, .high = addr + len - 1,
-    .space = space, .callback = callback };
+    .space = space, .skip_ref_flag = skip_ref_flag, .callback = callback };
```
shinezyy, line 29: similarly, .skip_ref_flag should be renamed to mmio_diff_type

shinezyy (changes requested):

include/device/map.h
xyyy1420: aleady fixed

src/device/io/map.c
poemonsense: I think 0 is correct. flag can include multiple flags. right?

include/device/map.h
xyyy1420: wip

include/device/map.h
```diff
@@ -38,17 +46,16 @@ static inline int find_mapid_by_addr(IOMap *maps, int size, paddr_t addr) {
...
   for (i = 0; i < size; i ++) {
     if (map_inside(maps + i, addr)) {
-      difftest_skip_ref();
       return i;
     }
   }
   return -1;
 }
 
 void add_pio_map(const char *name, ioaddr_t addr,
```
poemonsense, line 55: I would prefer adding another function called `add_mmio_map_with_diff` with this `mmio_diff_flag`. Then this existing function does not need to be changed and existing code is also unchanged. Only these devices that we want diff should call the new function with the appropriate flags.

src/memory/paddr.c
```diff
@@ -19,6 +19,7 @@
 #include <memory/paddr.h>
 #include <memory/store_queue_wrapper.h>
 #include <memory/sparseram.h>
+#include <device/map.h>
```
poemonsense, line 22: Is this header include required? It seems other lines are unchanged.

poemonsense:
I propose using `feat(device)` for the title. Then there will be no duplicate words

include/device/map.h
xyyy1420: good idea, i'll do it

src/memory/paddr.c
xyyy1420: i'll delete it

src/device/io/map.c
xyyy1420: > I think 0 is correct. flag can include multiple flags. right?

yes, i'll replace 0 with SKIP_FREE

include/device/map.h
xyyy1420: I added the `add_mmio_map_with_diff` api

cebarobot (commented):
I didn't see any codes that uses `mmio_diff_type`. Are they in the next PR? Also, Could you tell more about the background? Why we need selective difftest of mmio address space?

By the way, Port IO is only used in x86, which is actually useless now. But it's always good to provide those codes.

Current source codes LGTM.

xyyy1420:
> I didn't see any codes that uses `mmio_diff_type`. Are they in the next PR? Also, Could you tell more about the background? Why we need selective difftest of mmio address space?
> 
> By the way, Port IO is only used in x86, which is actually useless now. But it's always good to provide those codes.
> 
> Current source codes LGTM.

no code is using `MMIO_DIFF_TYPE`, but it is using elements defined in this type;
i want put the register state and gcpt into `flash`, this requires `flash` as a device that can properly use difftest, so i think we need to selective difftest of mmio address space

shinezyy (approved):
lgtm

cebarobot (approved):
LGTM

poemonsense (approved):
Some minor style improvements

Others look good to me

xyyy1420:
> Some minor style improvements
> 
> Others look good to me

already fixed

PR #745, opened 2024-12-31
feat(difftest): support difftest start from flash

xyyy1420:
* Bump ready-to-run
* delete `--restore` arg, we donot need it now, nemu could load checkpoint image directly
* modidy riscv64-nutshell-diff-spike_defconfig, support boot from flash and difftest start from flash
* refact image load in mnitor.c
* add pmp and pmp cpy api into difftest_attach, and fix some non-register state set issues for spike, but still could not make difftest_attach work


xyyy1420:
WIP

src/device/device.c
```diff
@@ -99,10 +99,6 @@ void init_device() {
   IFDEF(CONFIG_HAS_DISK, init_disk());
   IFDEF(CONFIG_HAS_SDCARD, init_sdcard());
   IFDEF(CONFIG_HAS_FLASH, init_flash());
-#ifndef CONFIG_SHARE
-  IFDEF(CONFIG_HAS_FLASH, load_flash_contents(CONFIG_FLASH_IMG_PATH));
-#endif // CONFIG_SHARE
```
poemonsense, line 104: Maybe we should seperate these flash.c, device.c changes into another PR for review.

This PR still seems too large for us to understand what's going on here

xyyy1420: ok, i'll split this pr

src/cpu/difftest/dut.c
```diff
@@ -103,13 +107,28 @@ void init_difftest(char *ref_so_file, long img_size, int port) {
   assert(ref_difftest_store_commit);
 #endif
 
+// prepare for difftest attach
+//  ref_difftest_pmpcpy = dlsym(handle, "difftest_pmpcpy");
+//  assert(ref_difftest_pmpcpy);
+//
+//  ref_difftest_pmp_cfg_cpy = dlsym(handle, "difftest_pmp_cfg_cpy");
+//  assert(ref_difftest_pmp_cfg_cpy);
```
cebarobot, line 115: Is this comment usefull?

src/cpu/difftest/dut.c
```diff
@@ -160,18 +179,28 @@ void difftest_step(vaddr_t pc, vaddr_t npc) {
   // Log("run ref %lx, %lx, %ld", pc, ref_r.pc, cpu.v);
   checkregs(&ref_r, pc);
 }
+
 void difftest_detach() {
+  Log("Difftest detach now!");
```
cebarobot, line 184: Should this log always print out?

src/device/flash.c
```diff
@@ -17,28 +17,80 @@
...
 // See the notes at paddr.c:79 for why we fix the address here.
 static uint8_t *flash_base  = (uint8_t *)0xf0000000ul;
-static FILE *fp = NULL;
 
 static void flash_io_handler(uint32_t offset, int len, bool is_write) {
   Assert(!is_write, "write to flash is illegal");
   return;
 }
 
+void convert_to_absolute_path(const char *input_path, char *output_path) {
```
cebarobot, line 32: I don't understand what this function for? Path problem should be handled by system, not NEMU.

By the way, flash load is similar to memory image load. The memory initialization does not use this.

cebarobot (commented):
If I understand properly, this PR allows loading flash from the command line? It seems that in previous version, the flash image path is hardcoded at compile time, but the flash image itself is not included in the finial executable file, which is weird.

I comletely agree with @poemonsense's opinion that this PR should be spilted. Every point in description may require a seperate PR.

src/device/flash.c
xyyy1420: > I don't understand what this function for? Path problem should be handled by system, not NEMU.
> 
> By the way, flash load is similar to memory image load. The memory initialization does not use this.

because `load_flash_contents` func using path from menuconfig, when we set "CONFIG_FLAH_IMG_PATH=$(NEMU_HOME)/resource/gcpt_restore/build/gcpt.bin", the system cannot handle it
`
void init_flash() {
#ifndef CONFIG_SHARE
  IFDEF(CONFIG_HAS_FLASH, load_flash_contents(CONFIG_FLASH_IMG_PATH));
#endif // CONFIG_SHARE
  add_mmio_map_with_diff("flash", CONFIG_FLASH_START_ADDR, flash_base, CONFIG_FLASH_SIZE, SKIP_FREE, flash_io_handler);
}
`

xyyy1420:
already spilt into

PR #747, opened 2025-01-01
opt(compile): allow aggressive inline

eastonman:
this allows get_effective_address() and
isa_mmu_check() to be inlined.
this improves about 10% of speed on zen4


Makefile
```diff
@@ -85,6 +85,12 @@ CFLAGS_BUILD += $(call remove_quote,$(CONFIG_CC_OPT))
 CFLAGS_BUILD += $(if $(CONFIG_CC_LTO),-flto=auto,)
 CFLAGS_BUILD += $(if $(CONFIG_CC_DEBUG),-ggdb3,)
 CFLAGS_BUILD += $(if $(CONFIG_CC_ASAN),-fsanitize=address,)
+
+# Fine-tuned compiler options
+ifeq ($(CC), gcc)
+	CFLAGS_BUILD += --param max-inline-insns-single=256 -falign-labels=32:9:64:15
```
cebarobot, line 91: I think it's better to have a option in Kconfig, as there is already some options about compilation?

poemonsense:
Does this PR still work?

eastonman:
Not sure on latest code. Will try.

PR #753, opened 2025-01-02
fix(args): delete restore arg, NEMU could load checkpoint image directly, we donot need this arg now

xyyy1420:


shinezyy:
LGTM.

Maybe add some info in commit message body:

Because NEMU can infer the type of input (whether it's image or compressed image), we don't need to manually specify `--restore`

shinezyy (approved):

PR #755, opened 2025-01-02
feat(flash): support load flash from cmd line, simply wrapper memory data initialize using fill_memory

xyyy1420:


poemonsense (approved):
LGTM. This is a good feature

shinezyy (approved):
Good job!

PR #757, opened 2025-01-03
try-fix(difftest attach): Add 'pmp' and 'pmp_cpy' API to difftest_attach, but functionality remains incomplete

xyyy1420:
this patch add two api to copy pmp and pmp cfg registers to spike when execute `difftest_attach`, but spike still have many non-register states not sync totally, so this patch just reduce size of the problem, not fix it


src/cpu/difftest/dut.c
```diff
@@ -179,6 +187,12 @@ void difftest_attach() {
   skip_dut_nr_instr = 0;
 
   isa_difftest_attach();
+
+  // There are multiple non-register states for spike,
+  // and entering attach from any execution phase requires
+  // the ability to correctly implement the settings for
+  // these states, and the expected repair effort would be enormous
+  assert(0); // fix me
```
shinezyy, line 195: // This function remains non-functional even after applying this patch, so we assert it.
// Spike has multiple non-register states, and to ensure correctness when attaching 
// from any execution phase, all these non-register states need to be synchronized.
// Addressing this properly in the future could require significant effort.

xyyy1420: already fixed

shinezyy (approved):

PR #762, opened 2025-01-06
feat: support using spike-so from spike CI

xyyy1420:
This patch add a CI job to download spike-so from OpenXiangShan/riscv-isa-sim CI, i think this patch could avoid bump spike in ready-to-run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,8 @@ on:
 
 env:
   CI_WORKLOADS: /nfs/home/share/ci-workloads
-  SPIKE_SO: ./ready-to-run/riscv64-spike-so
+  TMP_READY_TO_RUN: /nfs/home/share/ci-workloads/tmp_ready_to_run
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   compilation-all:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -27,17 +28,135 @@ jobs:
...
+      spike-so: ${{ steps.save-spike-so.outputs.SPIKE_SO }}
+      nutshell-spike-so: ${{ steps.save-nutshell-spike-so.outputs.NUTSHELL_SPIKE_SO }}
+    steps:
+    - name: Install jq
+      run: |
+          proxychains curl -L -o jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+          chmod +x ./jq
+          ./jq --version
+
+    # Step 1: 获取目标仓库最新运行的 workflow ID
```
shinezyy, line 44: plz change comments to English

.github/workflows/ci.yml
```diff
@@ -27,17 +28,135 @@ jobs:
...
+    name: Download spike-sp from CI
+    outputs:
+      spike-so: ${{ steps.save-spike-so.outputs.SPIKE_SO }}
+      nutshell-spike-so: ${{ steps.save-nutshell-spike-so.outputs.NUTSHELL_SPIKE_SO }}
+    steps:
+    - name: Install jq
+      run: |
+          proxychains curl -L -o jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+          chmod +x ./jq
+          ./jq --version
```
poemonsense, line 42: Is this `./jq --version` required?

.github/workflows/ci.yml
```diff
@@ -289,3 +422,28 @@ jobs:
...
+    steps:
+    - name: Setup env
+      run: |
+        echo "SPIKE_SO=${{ needs.download_spike_so.outputs.spike-so }}" >> $GITHUB_ENV
+        echo "NUTSHELL_SPIKE_SO=${{ needs.download_spike_so.outputs.nutshell-spike-so }}" >> $GITHUB_ENV
+    - name: Clean up
+      run: |
+        echo $(dirname $SPIKE_SO)
+        echo $(dirname $NUTSHELL_SPIKE_SO)
+        rm -rf $(dirname $SPIKE_SO)
```
poemonsense, line 448: If multiple CI runs simultaneously, will these `rm -rf` commands corrupt other CI runs?

xyyy1420: i think these command will not corrupt other CI runs, i add timestamp to `dirname $SPIKE_SO` and `dirname $NUTSHELL_SPIKE_SO`, so it just will clean up this CI run

.github/workflows/ci.yml
xyyy1420: could be deleted, just checking availability

.github/workflows/ci.yml
xyyy1420: image
for example, `$SPIKE_SO` and `$NUTSHELL_SPIKE_SO` looks like this

.github/workflows/ci.yml
xyyy1420: already fixed

.github/workflows/ci.yml
xyyy1420: already deleted

.github/workflows/ci.yml
poemonsense: Look good

.github/workflows/ci.yml
```diff
@@ -27,17 +28,134 @@ jobs:
...
+
+    # Step 1: get target repo latest workflow ID
+    - name: Get latest workflow run ID
+      id: get_workflow_run
+      run: |
+        REPO="OpenXiangShan/riscv-isa-sim"
+        WORKFLOW_NAME="build-difftest-so.yml"
+        RUNS_API="https://api.github.com/repos/$REPO/actions/workflows/$WORKFLOW_NAME/runs"
+
+        echo "Fetching latest workflow run ID from $RUNS_API..."
```
poemonsense, line 51: If it always downloads from the latest workflow, will it check whether it is from the master branch of Spike?

poemonsense (commented):
@cebarobot Take a look at this?

The other thing I'm worried is that, actions like this will bring dependence issues on the repos and may bring circular dependence in the future. For example, if there's one day Spike's CI depends on NEMU, then NEMU change will depend on Spike, and Spike depends on NEMU. Both cannot be merged if CI fails.

Similar issues exist in other OpenXiangShan repo, such as difftest. Currently I'm running only a nightly test for XiangShan to avoid the circular dependence issue. However, I think this is not a good idea.

We may need to think of a correct CI organization for our repos.

.github/workflows/ci.yml
xyyy1420: it could always get the specified branch and run status, i'll add these condition

xyyy1420: maybe we can specified the `head_sha` of Spike, it could solves the dependence problem, but there is still the problem of manually bump the Spike in CI

cebarobot (commented):
I have similar concerns as @poemonsense. I don't think that always downloading latest artifacts is a good idea for some reasons:

* There is version dependency problem between NEMU and SPIKE. ready-to-run keeps that dependency info, but this way will lose it.
* Running CI for multiple times should get the same result. This way will broke this.
* GitHub workflow artifacts have expiration time. They won't be keept too long.

We do need some better way to bump ready-to-run or keep the dependency info. Although downloading spike-so from artifacts is a simple and time-saving way, I am still very worried that it may cause more problems.

xyyy1420:
ok, there are still a lot of unresolved issues, this pr is closed for now, thank you to all reviewers

poemonsense:
> ok, there are still a lot of unresolved issues, this pr is closed for now, thank you to all reviewers

If you have free time, please help survey how other projects, especially modern open-source software projects, handle similar CI issues. There should be a lot of dependence issues in software developing. If you find some good methods to deal with these issues, we can refactor the CI for all OpenXiangShan repos. Thank you for your time

PR #766, opened 2025-01-07
refactor(Makefile): Migrate builds from other repos to separate makefiles

xyyy1420:
Simply migrate the other repo's build process to a separate makefile, and will continue to be refact to make it easier to understand.
The goal is to:
make it easier to handle builds and dependencies when there are more repositories, e.g. the Libcheckpoint repo and the nanopb repo will be added soon.

### DIFF
--- a/scripts/repos.mk
+++ b/scripts/repos.mk
@@ -1,0 +1,4 @@
+include $(NEMU_HOME)/scripts/softfloat.mk
+include $(NEMU_HOME)/scripts/LibcheckpointAlpha.mk
+
+clean-repos: clean-softfloat
--- a/scripts/repos.mk
+++ b/scripts/repos.mk
@@ -1,4 +1,4 @@
 include $(NEMU_HOME)/scripts/softfloat.mk
 include $(NEMU_HOME)/scripts/LibcheckpointAlpha.mk
 
-clean-repos: clean-softfloat
+clean-repos: clean-softfloat clean-libcheckpointalpha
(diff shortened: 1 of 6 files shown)


scripts/repos.mk
```diff
@@ -0,0 +1,4 @@
+include $(NEMU_HOME)/scripts/softfloat.mk
+include $(NEMU_HOME)/scripts/LibcheckpointAlpha.mk
+
+clean-repos: clean-softfloat
```
shinezyy, line 4: why don't clean libcheckpointalpha when clean-repos?

shinezyy:
Did you test the dependency manually?

- build from clean space
- build; delete softfloat or libcheckpoint alpha; build again

poemonsense (approved):
LGTM.

I notice that without riscv toolchain NEMU could not be built. This seems unnecessary because most NEMU users do not compile libcheckpoint. Maybe we should think of how we can avoid this and only compile libcheckpoint when necessary. This can be done in a future PR

scripts/repos.mk
xyyy1420: already fixed

xyyy1420:
> LGTM.
> 
> I notice that without riscv toolchain NEMU could not be built. This seems unnecessary because most NEMU users do not compile libcheckpoint. Maybe we should think of how we can avoid this and only compile libcheckpoint when necessary. This can be done in a future PR

ok, thanks for review, i'll fix it later

xyyy1420:
> Did you test the dependency manually?
> 
> * build from clean space
> * build; delete softfloat or libcheckpoint alpha; build again

yes, i test manually, looks like working

shinezyy (approved):

PR #771, opened 2025-01-08
fix: split LibcheckpointAlpha build process from menuconfig

xyyy1420:
follow up to #766;

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,15 +34,13 @@ CFLAGS += -D__MMIO_SPECE_RANGE__=$(call remove_quote,$(CONFIG_MMIO_SPACE_RANGE))
 endif
 
 ifdef CONFIG_HAS_FLASH
-
 FLASH_IMG_PATH=$(shell realpath $(CONFIG_FLASH_IMG_PATH) 2>/dev/null)
 ifeq ($(FLASH_IMG_PATH),)
 CFLAGS += -D__FLASH_IMG_PATH__=\"\"
 else
 CFLAGS += -D__FLASH_IMG_PATH__=\"$(FLASH_IMG_PATH)\"
-endif
-
-endif
+endif # ($(FLASH_IMG_PATH),)
+endif # CONFIG_HAS_FLASH
 
 INC_DIR += $(NEMU_HOME)/src/isa/$(ISA)/include
 DIRS-y += src/isa/$(ISA)
--- a/Makefile
+++ b/Makefile
@@ -34,12 +34,14 @@ CFLAGS += -D__MMIO_SPECE_RANGE__=$(call remove_quote,$(CONFIG_MMIO_SPACE_RANGE))
 endif
 
 ifdef CONFIG_HAS_FLASH
+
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -34,12 +34,14 @@ CFLAGS += -D__MMIO_SPECE_RANGE__=$(call remove_quote,$(CONFIG_MMIO_SPACE_RANGE))
 endif
 
 ifdef CONFIG_HAS_FLASH
+
```
poemonsense, line 37: unused empty lines

poemonsense (approved):
The change looks good to me. So now the user must explicitly build the libcheckpoint `make gcpt_restore_bin` before creating chekcpoints. We may need to add this command to CI / user doc

Makefile
```diff
@@ -34,12 +34,14 @@ CFLAGS += -D__MMIO_SPECE_RANGE__=$(call remove_quote,$(CONFIG_MMIO_SPACE_RANGE))
...
 FLASH_IMG_PATH=$(shell realpath $(CONFIG_FLASH_IMG_PATH) 2>/dev/null)
 ifeq ($(FLASH_IMG_PATH),)
 CFLAGS += -D__FLASH_IMG_PATH__=\"\"
 else
 CFLAGS += -D__FLASH_IMG_PATH__=\"$(FLASH_IMG_PATH)\"
 endif
+
```
shinezyy, line 44: feel like you added spaces here to make the code clearer, right?
Would it be better to change it like this?

endif // ($(FLASH_IMG_PATH),)  

endif  // CONFIG_HAS_FLASH

scripts/LibcheckpointAlpha.mk
```diff
@@ -1,16 +1,21 @@
...
+ifeq ($(wildcard $(CPT_MEMLAYOUT_HEADER)),)
+  $(shell git clone --depth=1 https://github.com/OpenXiangShan/LibCheckpointAlpha.git $(CPT_REPO))
+endif
+
+gcpt_restore_bin: $(CPT_BIN)
 
-$(CPT_BIN): $(CPT_LAYOUT_HEADER)
-	$(Q)$(MAKE) $(silent) -C $(LIB_CPT_PATH) CROSS_COMPILE=$(CPT_CROSS_COMPILE)
+$(CPT_BIN):
+ifdef CONFIG_HAS_FLASH
```
poemonsense, line 14: Given that this is not in the dependence any more, maybe we don't need to check CONFIG_HAS_FLASH here

shinezyy (commented):
Looks much cleaner!

Makefile
xyyy1420: already fixed

scripts/LibcheckpointAlpha.mk
xyyy1420: yes, i delete these `ifdef`

Makefile
xyyy1420: already fixed

shinezyy (approved):

PR #773, opened 2025-01-09
feat: add Libcheckpoint repo

xyyy1420:
prepare replace LibcheckpointAlpha with LibCheckpoint


shinezyy:
Since libcheckpoint is not currently being used, how can we verify that this patch is correct?
Should we also include CI in thsi patch?

xyyy1420:
> Since libcheckpoint is not currently being used, how can we verify that this patch is correct? Should we also include CI in thsi patch?

Libcheckpoint is not currently being used, will not broken any NEMU function, just clone repo, i think we can add CI when we use Libcheckpoint?

shinezyy (approved):
ok

PR #778, opened 2025-01-14
feat: add --store-cpt-in-flash option

xyyy1420:
This commit introduces the `--using-flash-store-checkpoint` option, which is intended to enable saving checkpoints directly to flash storage.

Currently, this option is not yet implemented in the workflow. Future updates will integrate this feature into the relevant checkpoint handling logic.

### DIFF
--- a/include/checkpoint/serializer.h
+++ b/include/checkpoint/serializer.h
@@ -36,7 +36,7 @@ class Serializer
 
     explicit Serializer();
 
-    void init();
+    void init(bool flash_store_checkpoint);
 
     bool shouldTakeCpt(uint64_t num_insts);
     bool instrsCouldTakeCpt(uint64_t num_insts);
@@ -68,6 +68,7 @@ class Serializer
     const uint32_t MISCDoneFlag;
 
     bool regDumped{false};
+    bool flash_store_checkpoint{false};
 
     std::map<uint64_t, double> simpoint2Weights;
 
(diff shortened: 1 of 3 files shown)


include/checkpoint/serializer.h
```diff
@@ -36,7 +36,7 @@ class Serializer
 
     explicit Serializer();
 
-    void init();
+    void init(bool flash_store_checkpoint);
```
shinezyy, line 39: I prefer store_checkpoint_in_flash, store_cpt_in_flash, or cpt_in_flash as var name

src/monitor/monitor.c
```diff
@@ -115,6 +116,7 @@ static inline int parse_args(int argc, char *argv[]) {
     {"cpt-mmode"          , no_argument      , NULL, 7},
     {"map-cpt"            , required_argument, NULL, 10},
     {"checkpoint-format"  , required_argument, NULL, 12},
+    {"using-flash-store-checkpoint", no_argument, NULL, 17},
```
shinezyy, line 119: similarly, the option name can be store_checkpoint_in_flash or store_cpt_in_flash

shinezyy (changes requested):

include/checkpoint/serializer.h
xyyy1420: already fixed

src/monitor/monitor.c
xyyy1420: already fixed

shinezyy (approved):

PR #780, opened 2025-01-14
feat: replace fixed get_pmem with serialize_base_addr for checkpoint storage

xyyy1420:
Replaced the hardcoded `get_pmem` with `serialize_base_addr` to enable specifying the checkpoint storage location via the `--using-flash-store-checkpoint` option. While this change lays the groundwork for saving checkpoints to flash, the functionality is not yet fully implemented and will be completed in future updates.


shinezyy (approved):
lgtm

PR #783, opened 2025-01-15
refactor(checkpoint): refactor checkpoint file naming logic

xyyy1420:


src/checkpoint/serializer.cpp
```diff
@@ -122,39 +124,52 @@ void Serializer::serializePMem(uint64_t inst_count) {
                     ? numeric_limits<int>::max()
                     : ((int64_t)PMEM_SIZE - (int64_t)written);
 
-      if (gzwrite(compressed_mem, pmem + written, (uint32_t)pass_size) != (int)pass_size) {
+      if (gzwrite(memory_compressed_mem, pmem + written, (uint32_t)pass_size) != (int)pass_size) {
```
shinezyy, line 127: Is memory_compressed_mem better than compressed_mem? Why?
Is memory_compressed_file ok?

xyyy1420: memory_compressed_file is ok, i'll fix it
and i'll add flash_compressed_file in next pr

shinezyy (approved):

PR #803, opened 2025-02-09
fix(priv,exception): Raise exception for unsupported priv-op if REPORT_ILLEGAL_INSTR was disabled

forever043:


poemonsense (approved):

forever043:
included in #804

PR #811, opened 2025-02-17
feat(smcntrpmf): support Smcntrpmf extension

wissygh:


src/isa/riscv64/system/priv.c
```diff
@@ -882,39 +890,46 @@ inline word_t gen_status_sd(word_t status) {
   return ((word_t)(fs_dirty || vs_dirty)) << 63;
 }
 
+#ifdef CONFIG_RV_SMCNTRPMF
+static inline bool inbibit_smcntrpmf(word_t current_cfg) {
```
sinceforYy: why inbibit?

PR #823, opened 2025-03-07
feat(semantic point): support semantic checkpoint

xyyy1420:
A semantic checkpoint identifies checkpoint positions based on a set of basic blocks and their execution counts. The first instruction after a branch instruction is considered the input of a basic block, while the branch instruction itself serves as the output of the basic block.

Although SimPoint can identify hotspot locations in program execution, it can only mark checkpoint positions based on the number of executed instructions. However, many factors can influence the instruction count, making this identification potentially inaccurate. As a result, generating multiple checkpoint based on SimPoint's results may lead to checkpoints that actually belong to different phases of the program.

Semantic checkpoint identifies program execution positions based on a series of basic block execution counts, ensuring that:

  1. The execution path remains the same (covering the given sequence of basic blocks).
  
  2. The checkpoint positions remain consistent across multiple generations (the basic block sequence counters match the given values).


xyyy1420:
```
0x1cf90,0x1cf98,14771
0x1cf9c,0x1cfac,14772
0x1cfd6,0x1cfe6,14773
0x1d0f6,0x1d0fe,14784
0x1d102,0x1d112,14785
0x1d066,0x1d07c,14793
0x1d0b6,0x1d0be,14794
0x1cef6,0x1cf14,14795
0x1cf4e,0x1cf56,14796
0x1d020,0x1d028,14797
0x1d114,0x1d132,14798
0x1f070,0x1f088,15054
0x1dd66,0x1dd94,15280
0x1dd98,0x1ddb6,15281
0x1ddde,0x1dde8,15286
0x1ddea,0x1de14,15287
0x1de18,0x1de3e,15288
0x1ddc2,0x1ddda,15291
0x1de42,0x1de48,15292
0x1dd36,0x1dd62,15305
```
为astar_rivers生成checkpoint的参考输，每一行分别是basic_block_first_pc,basic_block_second_pc,exec_times

shinezyy (changes requested):
1. typo: semantic instead of sematic. The file name should  also be corrected
2. why c++ 20? is it necessray? What is the minimal gcc/clang version required? Will it fail to compile on some old servers?

xyyy1420:
> 1. typo: semantic instead of sematic. The file name should  also be corrected
> 2. why c++ 20? is it necessray? What is the minimal gcc/clang version required? Will it fail to compile on some old servers?

1. i'll fix it
2. C++20 seems to support hash value computation for `pair structures`, which can be used in `unordered_map`. However, our server does not yet have a compiler that supports C++20, so this feature is not actually being used. Moreover, it does not cause any compilation errors. i'll delete it

shinezyy (approved):

PR #833, opened 2025-03-21
fix(pbmt): add cpu state for amo and misalign check

Maxpicca-Li:


Anzooooo:
Why does only `read` need to check `nc`? :eyes:

Maxpicca-Li:
> Why does only `read` need to check `nc`? 👀

thanks! I have already fixed.

Anzooooo (approved):

PR #834, opened 2025-03-24
chore: fix so generation

Maxpicca-Li:
`riscv64-xs-ref_defconfig` is currently not being generated properly, so here need `make clean-all`.


cebarobot (approved):
LGTM

PR #850, opened 2025-04-15
fix(pbmt): fix misalign write fault

Anzooooo:


src/memory/vaddr.c
```diff
@@ -38,7 +38,7 @@ static paddr_t vaddr_trans_and_check_exception(vaddr_t vaddr, int len, int type,
     return 0;
   }
   if (cpu.pbmt != 0) {
-    isa_mmio_misalign_data_addr_check(paddr, vaddr, len, MEM_TYPE_READ, true);
+    isa_mmio_misalign_data_addr_check(paddr, vaddr, len, MEM_TYPE_WRITE, true);
```
Maxpicca-Li, line 41: It is better to change `MEM_TYPE_WRITE` to `type` here. If the `type` passed in is `MEM_TYPE_READ`, an error will occur here.

Maxpicca-Li (changes requested):

PR #859, opened 2025-04-26
refactor(priv): not use slow path for priv instructions

cebarobot:
The slow path handler for privileged instructions was introduced in commit e6b9212c7153a273b0ca3000ab1a711718af3718 (dated March 18, 2021). Previously, the slow path was beneficial in `PERF_OPT` mode: a bundle of rarely used instructions was handled as a single instruction type, reducing both the number of `EHelper` (Execute Helper) functions and the size of the `execute()` function.

However, decoding specific instructions had to be handled within the slow path `EHelper`, which could not benefit from the regular decoding flow (such as using patterns like `def_INSTR_TAB("0011000 00010 00000 ??? 00000 ????? ??", mret);`). This introduced additional encoding complexity and reduced the readability and maintainability of the code.

This PR removes the `def_SYS_EHelper` macro and the `rtl_sys_slow_path()` function. It also refactors `priv_instr()` and `csrrw()` by splitting them into several small functions, each corresponding to a specific instruction (with names prefixed by `riscv64_priv_`). Additionally, it restructures the handling logic of `hload` and `hstore`. As a result, executing privileged instructions no longer requires the complicated call chain `rtl_sys_slow_path() -> rtl_hostcall() -> isa_hostcall() -> priv_instr()`.

This PR compares the performance between the slow path (previous implementation) and the normal path (current implementation). We observed that:
* With `PERF_OPT` enabled, the execution time of `linux-hello` and `xvisor-hello` increases by 0.68%;
* With `PERF_OPT` disabled, the execution time of `linux-hello` decreases by 0.8%.

This PR includes 4 commits, and is recommanded to be merged into master by rebase:
* refactor(priv): split priv_instr() into several functions
* refactor(fence.i): set flush flag directly
* refactor(mop): simplify the decode and exec helper
* refactor(priv): totally abandon rtl_sys_slow_path


src/isa/riscv64/instr/rvc/decode.h
```diff
@@ -503,9 +503,7 @@ def_THelper(rvc_Q2_jr_mov) {
 }
 
 def_THelper(rvc_Q2_jalr_add) {
-#if defined(CONFIG_RV_DEBUG) || defined(CONFIG_EBREAK_AS_TRAP)
```
poemonsense, line 506: why remove these macros

src/isa/riscv64/instr/rvc/exec.h
```diff
@@ -62,15 +62,15 @@ def_EHelper(c_jalr) {
 }
 
 def_EHelper(c_ebreak) {
-#ifdef CONFIG_RV_DEBUG
```
poemonsense, line 65: why remove this macro

poemonsense (commented):
It's difficult for me to review the functionalities of the code.

Do we consider introducing c++ formatter in NEMU?

src/isa/riscv64/instr/rvc/decode.h
cebarobot: Because EBREAK/C.EBREAK is a normal RVI/RVC instruction in the latest unpriv spec, which should be always implemented.

I wonder whether there should be a switch for compatibility?

src/isa/riscv64/instr/rvc/exec.h
cebarobot: Because EBREAK/C.EBREAK is a normal RVI/RVC instruction in the latest unpriv spec, which should be always implemented.

I wonder whether there should be a switch for compatibility?

poemonsense (approved):
Maybe we can merge first and fix bugs if any

PR #867, opened 2025-05-06
fix(paddr): Show detailed info when check_paddr failed

forever043:


cebarobot (approved):
LGTM. It's better if there is a switch for this log.

PR #889, opened 2025-05-22
config: add RVV_UNORDER option to control the unorder rvv insts

tastynoob:
Just add one independent option to switch the rvv unorder behavior.


Ziyue-Zhang:
Some config file alse need to be changed, such as spike diff config

poemonsense (commented):
Need to bump the configs

Issue #927, opened 2025-09-13
Not All `tcache` Related Function Declarations and Calls Controlled by the Macro `CONFIG_PERF_OPT`


scratch-er:
I have noticed that in `src/cpu/tcache.c`, there are two sets of functions. Which set is enabled is controlled by the macro definition `CONFIG_PERF_OPT`. The only place where these functions are involved is `src/cpu/cpu-exec.c`. These functions are related with decoding caching. 

However, not all of the function calls and declarations in `src/cpu/cpu-exec.c` are controlled by the macro definition. This can make the declaration, the function call and the actual definition inconsistent under some configurations. And this can cause compiler errors, especially when building the differential testing dynamic libraries using the configurations under:

https://github.com/OpenXiangShan/NEMU/tree/master/configs

scratch-er:
This also makes the CI fail in container `ghcr.io/openxiangshan/xs-env:latest`.

[full_act_log.txt](https://github.com/user-attachments/files/22311245/full_act_log.txt)



CI script:

```yml
  compilation-all:
    runs-on: ubuntu-latest
    container:
      image: ghcr.io/openxiangshan/xs-env:latest
    continue-on-error: false
    name: Compilation - all configs
    timeout-minutes: 60
    steps:
    - name: Checkout repository
      uses: actions/checkout@v4
    - name: Setup env
      run: |
        echo "NEMU_HOME=$GITHUB_WORKSPACE" >> $GITHUB_ENV
    - name: Try to compile
      id: list_configs
      run: |
        bash ./scripts/compilation-test.sh
```

log:

```
[CI/Compilation - all configs]   | src/cpu/cpu-exec.c:848:26: error: passing arguments to 'tcache_handle_flush' without a prototype is deprecated in all versions of C and is not supported in C23 [-Werror,-Wdeprecated-non-prototype]
[CI/Compilation - all configs]   |   848 |       tcache_handle_flush(cpu.pc);
[CI/Compilation - all configs]   |       |                          ^
[CI/Compilation - all configs]   | src/cpu/cpu-exec.c:877:28: error: passing arguments to 'tcache_handle_flush' without a prototype is deprecated in all versions of C and is not supported in C23 [-Werror,-Wdeprecated-non-prototype]
[CI/Compilation - all configs]   |   877 |         tcache_handle_flush(cpu.pc);
[CI/Compilation - all configs]   |       |                            ^
[CI/Compilation - all configs]   | 2 errors generated.
```

[...]

PR #929, opened 2025-09-14
fix(clang): guard all tcache functions with macro (#927)

scratch-er:
Resolve compilation errors in clang by ensuring all tcache related function calls in src/cpu/cpu-exec.c are properly guarded by the `CONFIG_PERF_OPT` macro.

Some of these function calls were not guarded with macro. This caused inconsistency between declarations, calls, and actual definitions, resulting in C23 compatibility errors when building with clang.

### DIFF
--- a/src/cpu/cpu-exec.c
+++ b/src/cpu/cpu-exec.c
@@ -502,7 +502,7 @@ end_of_loop:
   debug_difftest(this_s, s);
   save_globals(s);
 }
-#else
+#else // CONFIG_PERF_OPT
 #define FILL_EXEC_TABLE(name) [concat(EXEC_ID_, name)] = concat(exec_, name),
 
 #define rtl_priv_next(s)
@@ -731,7 +731,7 @@ static void execute(int n) {
   }
   Loge("total insts: %'lu, execute remain: %'d", get_abs_instr_count(), n_remain);
 }
-#endif
+#endif // CONFIG_PERF_OPT
 
 IFDEF(CONFIG_DEBUG, char log_bytebuf[80] = {};)
 // max size is (strlen(str(instr)) + strlen(suffix_char(id_dest->width)) + sizeof(id_dest->str) + sizeof(id_src2->str) + sizeof(id_src1->str))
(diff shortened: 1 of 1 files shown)


Anzooooo (commented):
I have no issues with the task itself.
However, there are some other related matters. For tcache, there may be additional code logic not fully controlled by CONFIG_PERF_OPT, such as:
https://github.com/OpenXiangShan/NEMU/blob/dee3a5ecfa159df4a386765ece47d102d98aad96/src/cpu/cpu-exec.c#L685-L698
These should not cause compilation errors, but they should not occur when CONFIG_PERF_OPT is not defined.
Perhaps we should review all the logic and fully control it using CONFIG_PERF_OPT?

Anzooooo:
Yes, based on your output, when `CONFIG_PERF_OPT` is not defined, the code blocks above will be defined.
As ceba mentioned in issue 927(https://github.com/OpenXiangShan/NEMU/issues/927#issuecomment-3288492298), this may be an unreasonable behavior.

In fact, all this `tcache` related code was introduced by this PR(https://github.com/OpenXiangShan/NEMU/commit/c503e7a86d48fd560292a7e085795240fb3085df#diff-1f9d263325b615223a888631b8e514d3b1b7d6e3c4ea5220d67a29e2acaa9155R696).
This PR didn't add macro control for these code sections. Since you've added macro control for the parts introduced by this PR, perhaps you could consider adding macro control for all of them?
Also, as Ceba mentioned earlier, do we need an additional macro to distinguish these code sections?

scratch-er:
The most direct problem affected by #927 is that it causes the compilation of the `riscv64-xs-clang-ref_defconfig` configuration to fail during the `compilation-all` task. Therefore, the code before my modification was likely problematic. Personally, I suspect it was an oversight by the author who introduced the `tcache` related code.

As for whether there is a better solution, I believe it requires discussion. After all, the macro definitions in `cpu-exec.c` have become so complex that they are somewhat difficult to manage.

Anzooooo:
> The most direct problem affected by #927 is that it causes the compilation of the `riscv64-xs-clang-ref_defconfig` configuration to fail during the `compilation-all` task, under `ghcr.io/openxiangshan/xs-env:ubuntu-24.04`. Therefore, the code before my modification was likely problematic. Personally, I suspect it was an oversight by the author who introduced the `tcache` related code.
> 
> As for whether there is a better solution, I believe it requires discussion. After all, the macro definitions in `cpu-exec.c` have become so complex that they are somewhat difficult to manage.

As I mentioned earlier, I have no reservations about this PR itself, which aims to fix the compilation errors in CI.

Regarding whether other sections should be wrapped in macros, we can revisit Ceba's opinion on that. @cebarobot

poemonsense (approved):
Sorry for the late reply.

This PR seems good to me. See https://github.com/OpenXiangShan/NEMU/issues/927#issuecomment-3565246499

@Anzooooo @good-circle @cebarobot Take a look when you have time? Thanks

PR #937, opened 2025-11-05
feat(pma,pmp): supoort 32 pmp and pma

sinceforYy:


src/isa/riscv64/Kconfig
```diff
@@ -309,6 +309,12 @@ config RV_PMP_ENTRY_64
   help
     All pmpcfg and pmpaddr CSRs are accessible.
 
+config RV_PMP_ENTRY_64_ACTIVE_32
+  bool "32 active PMP entries"
+  help
+    Only pmpcfg0, pmpcfg2, pmpcfg4, pmpcfg6, pmpaddr0 - pmpaddr31 are accessible.
+    pmpcfg8 - pmpcfg14, pmpaddr32 - pmpaddr63 are read-only zero.
```
cebarobot: I don't think this config is not useful. There is already a `RV_PMP_ACTIVE_NUM` here. If you want to have 64 PMP entries with 32 entries active, just set `RV_PMP_ACTIVE_NUM` to 32.

sinceforYy: Ok, I will delete it.

src/memory/paddr.c
```diff
@@ -40,7 +40,7 @@ bool need_read_golden_mem = false;
 #ifdef CONFIG_LIGHTQS
 #define PMEMBASE 0x1100000000ul
 #else
-#define PMEMBASE 0x700000000000UL
+#define PMEMBASE 0x70000000000UL
```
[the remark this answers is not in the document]
sinceforYy: #939

cebarobot (changes requested):
LGTM, but there are still two issues:

* Is it decided to use 32 PMP and 32 PMA entries as default in Kunminghu-v2r2?
* If it will be default, could you also modify the following defconfig:
  * `riscv64-xs-ref_defconfig`
  * `riscv64-xs-ref-debug_defconfig`
  * `riscv64-dual-xs-ref_defconfig`
  * `riscv64-dual-xs-ref-debug_defconfig`
  * `riscv64-xs-ref_bitmap_defconfig`
  * And all non-ref defconfigs of the above defconfigs.

cebarobot (approved):
LGTM

PR #938, opened 2025-11-05
fix(csr): fix set/get sip/sie in V mode

sinceforYy:
Guest read/write operations on sie/sip in VS mode are functionally equivalent to host read/write operations on vsie/vsip in M mode. 
vmode_set_sip and set_vsip, as well as vmode_set_sie and set_vsie, are not entirely equivalent. Therefore, replace them with the correct set_vsip and set_vsie methods.


sinceforYy:
Yes, this bug is caused by a misunderstanding of the RISC-V priv spec. Additionally, regarding PMEM, I will rebase it.

PR #939, opened 2025-11-06
fix(paddr): move PMEMBASE to avoid conflicts

sinceforYy:
This patch moves PMEMBASE to a address(0x70000000000UL, 7TB), which alleviates the conflict.


poemonsense:
See https://github.com/OpenXiangShan/NEMU/pull/930#issuecomment-3500945244

PR #940, opened 2025-11-07
fix(paddr): avoid fixed address for mmap memory

poemonsense:
Fixed address space brings conflicts and we remove it.

Initially (dedab9c8db4c7f2df87a9fc1fdb35d11c8946f82), pmem is a const pointer whose base address is the same as the guest physical memory. This setting avoids the extra guest-host address translation overhead because their physical addresses are the same.

However, #61 removes the const attribute (for what? don't know) and we also find that there may be address conflicts if we are using the fixed address (#930, #939).

Therefore, we cannot use fixed address any more and thus we cannot avoid the guest-host address translation overhead (because non-const cannot be optimized anyway).

We are now dynamically allocating the space and assign it to pmem. If pmem is set to nonzero, we assume it has been allocated. This is useful when init_mem is called multiple times when simulating multiple workloads.

### DIFF
--- a/src/memory/paddr.c
+++ b/src/memory/paddr.c
@@ -37,12 +37,6 @@ extern uint64_t vec_read_golden_mem_data;
 bool need_read_golden_mem = false;
 #endif // defined(CONFIG_MULTICORE_DIFF) && defined(CONFIG_RVV)
 
-#ifdef CONFIG_LIGHTQS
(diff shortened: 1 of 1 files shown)


cebarobot (approved):
This makes a lot of sense and solves the problem forever. LGTM

PR #944, opened 2025-11-23
feat(difftest): add configs for optional CSR states

poemonsense:
Some CSR states may not be compared or synchronized even if their corresponding extensions are enable. For example, the user may choose only to compare floating-point registers without fcsr when FPU is enabled. In such cases, we need avoid some states in the cpu struct.

We also add the difftest_init_v2 API to allow an extra input that specifies the state size. This new API will check whether DUT and NEMU have the same size of state. Note it's still hard to ensure that each register is in the same position.


poemonsense:
I just update the commit/PR title to explicitly note only CSRs can be optional

PR #948, opened 2025-11-25
feat(riscv64): add config for disabling Sv39 ASIDs

poemonsense:
Some CPUs may not implement ASIDs. Currently only Sv39 implements this.

### DIFF
--- a/src/isa/riscv64/Kconfig
+++ b/src/isa/riscv64/Kconfig
@@ -485,19 +485,19 @@ config RV_SV48
 endchoice
 
 config RV_ASID_LEN
-  int "The number of implemented ASID bits (ASIDLEN)"
+  int "Number of implemented ASID bits (ASIDLEN)"
   depends on RV_SV39 || RV_SV48
   help
-    The number of implemented ASID bits in satp / vsatp (ASIDLEN), between 0 and 16.
+    Number of implemented ASID bits in satp / vsatp (ASIDLEN), between 0 and 16.
     Unimplemented bits will be read-only zero.
   range 0 16
   default 16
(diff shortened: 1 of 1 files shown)


src/isa/riscv64/Kconfig
```diff
@@ -484,6 +484,13 @@ config RV_SV48
 
 endchoice
 
+config RV_SV39_NO_ASID
+  bool "No ASIDs (always read/write zero)"
+  depends on RV_SV39 && !RVH
+  help
+    The ASID field of CSR satp is always read/write zero.
+  default n
+
```
cebarobot: This is not a good way to implement ASID config. According to RISC-V Priv spec, the implemented ASID bits can be any number between 0 and 16.

cebarobot (changes requested):
Sorry, but I can't accept this PR, which introduces something not very good.

I'd like to give a better implementation of this feature.

src/isa/riscv64/Kconfig
poemonsense: I think we don't need to worry about non-existing things. If this feature is useless, we should not spend too much time on this. Just implement a minimum support is enough

cebarobot: I have to say, current implementation is a detour and very ugly. Actually, just add an config to replace marco `SATP_ASID_LEN` is enough. No need to add any more ifdefs.

cebarobot (approved):

poemonsense:
> @poemonsense Hi, I have finished the new version of codes, please check and review.

Thanks. Look very good to me

PR #949, opened 2025-11-25
ci: migrate most jobs into container `xs-env`

scratch-er:
The CI jobs `compilation-all`, `basic-xiangshan`, `basic-nutshell`, `diff-spike-basic` and `diff-spike-checkpoint` are migrated into container `xs-env`.

- Workloads are downloaded from the latest release URL of `workload-builder` and cached.
- `take_zstd.sh` is modified to accept workload and output path from command line, so some jobs can generate checkpoints on the fly.
- `compilation-all` now runs on a container matrix of `xs-env:ubuntu-24.04` and `xs-env:ubuntu-22.04`. Ubuntu 20.04 is too old to run this job.
- `basic-xiangshan` now runs in `xs-env:ubuntu-24.04`, with SPEC workloads replaced with `coremark-pro`.
- `basic-nutshell` now runs in `xs-env:ubuntu-24.04`. Differential testing is removed from this job due to CSR config mismatch between NEMU DUT and spike REF.
- `diff-spike-basic` now runs in `xs-env:ubuntu-24.04`, with some workloads removed. 2 tests in `riscv-vector-tests` are temporarily skipped because the newly built version fail.
- `diff-spike-checkpoint` now runs in `xs-env:ubuntu-24.04`, generating checkpoints on the fly, with SPEC workloads replaced with `coremark-pro` and `rvv-bench`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,22 +275,32 @@ jobs:
 
       - name: Basic - rvv-test
         env:
-          dir_tests: ${{ env.CI_WORKLOADS }}/am/riscv-vector-tests/bin
+          # These tests are included in the workloads on the `nemu` runner.
+          # But the newly built versions (built with `workload-builder`) of them fail on NEMU.
+          # So skip them temporarily.
+          skip_tests: |
+            vsetvli-0.bin
+            vsetivli-0.bin
         run: |
-          find ${{ env.dir_tests }} -type f -name "*.bin" | while read -r test_bin; do
-            echo ::group::$(basename $test_bin)
-            ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} ${test_bin}
-            echo "::endgroup::"
+          readarray -t skip_tests_array <<< "$skip_tests"
+          find "${CI_WORKLOADS}/am/riscv-vector-tests/bin" -type f -name "*.bin" | while read -r test_bin; do
+            if [[ " ${skip_tests_array[@]} " =~ " $(basename $test_bin) " ]]; then
+              echo "$test_bin skiped."
+            else
+              echo ::group::$(basename $test_bin)
+              ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} ${test_bin}
+              echo "::endgroup::"
+            fi
           done
 
       - name: Basic - misc-tests
-        # if: false
         env:
           tests: |
             am/misc-tests/bin/bitmanip.bin
             am/coremark/bin/coremark-riscv64-xs-rv64gc-o2.bin
             am/coremark/bin/coremark-riscv64-xs-rv64gc-o3.bin
             am/coremark/bin/coremark-riscv64-xs-rv64gcb-o3.bin
+            am/misc-tests/bin/amtest-riscv64-xs.bin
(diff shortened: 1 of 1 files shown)


scratch-er:
Seems that `compilation-all` is hard to port to `xs-env:ubuntu-20.04`. So I removed it from the version matrix of this job.

- The `riscv64-linux-gnu-gcc` in that container is too old to support some vector instructions, which is required by `libcheckpoint`.
- The `clang` is too old to support `-flto=auto`, which is required by `riscv64-xiangshan-clang_defconfig`.

This also means that `xs-env:ubuntu-20.04` is not suitable when developing RVV related features.

.github/workflows/ci.yml
```diff
@@ -213,98 +275,93 @@ jobs:
...
         run: |
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/hmmer_nph3_1886_0.000268086.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/hmmer_retro_6200_0.0378585.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/h264ref_sss_88321_0.0346343.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/h264ref_foreman.baseline_8028_0.0414445.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/h264ref_foreman.main_3027_0.0443573.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/libquantum_41028_0.0840681.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
-          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $TEST_HOME/bzip2_402_0.00785398.gz -r $TEST_HOME/v-gcpt.bin -I 40000000
+          ./build/riscv64-nemu-interpreter -b --diff ${SPIKE_SO} $(find checkpoints/coremark-pro/test/linux/50000000 -name "*.zstd") -I 200000000
```
poemonsense: Does the `Take checkpoint` generate only one test?

Maybe we need more tests here. However, we can list it as TODO now.

poemonsense (approved):
Overall looks good to me.

I think we need to add a TODO/FIXME list here which should at least include:

* xvisor has been removed (in basic-xiangshan and diff-spike-basic)
* 2 tests in rvv-test are skipped
* several misc-tests are skipped
* diff-spike-checkpoint uses less workloads than before

This list helps developers to track what this PR changes.

.github/workflows/ci.yml
scratch-er: Yes, only one test is generated.

The tests that are removed are from SPEC. The new test added is coremark-pro, which is an open source alternative of SPEC. Please tell me why are more tests needed and suggest some tests to add.

PR #950, opened 2025-11-28
feat(configs): add files for yanqihu and nahu

poemonsense:
We add the defconfig files for yanqihu and nanhu architectures of the XiangShan CPU.


cebarobot (approved):
LGTM

PR #962, opened 2025-12-09
ci: add instruction count performance measurement

cyyself:
This commit adds instruction count performance measurement capability to the NEMU CI workflow. It utilizes DynamoRIO's inscount tool to count the number of instructions executed during the test runs. This enhancement allow us to see the performance impact of our changes in terms of instruction count, helping to identify potential optimizations and performance regressions early in the development process.


cebarobot (approved):
LGTM

poemonsense:
We should add IPS (instructions executed per x86 second) to the report as an end-to-end performance indicator.

poemonsense (approved):

PR #963, opened 2025-12-11
feat(uartlite): implement UARTLite INPUT FIFO

cyyself:
Currently, even the CONFIG_UARTLITE_INPUT_FIFO option exists, it cannot successfully coompiled due to missing implementation. This commit adds the missing implementation to enable the INPUT FIFO feature of UARTLite device.

With this feature, we can directly type characters into the UARTLite serial console.


cebarobot (approved):
LGTM

Issue #366, opened 2024-06-19
Compilation Failure When Using clang and Enabling RVB & PERF_OPT


cebarobot:
clang is not happy with current RVB implementation when perfomance optimization, because:
* NEMU defines some functions in `src/isa/riscv64/instr/rvb/rvintrin.h`.
* `rvintrin.h` is included in file `src/isa/riscv64/instr/rvb/exec.h`.
* `rvb/exec.h` is included in file `src/isa/riscv64/include/isa-exec.h`.
* `isa-exec.h` is included in function `execute()` in file `src/cpu/cpu-exec.c`, when perfomance optimization is enabled.

So that there is some functions are defined in another function, which clang complained about.

Also, `rvintrin.h` is really weird. It has another copyright header. I think this file is from [here](https://github.com/riscv/riscv-bitmanip/blob/main-history/cproofs/rvintrin.h), but the two files are very different. 

I think we should rewrite the implementation of RVB.

PR #964, opened 2025-12-12
feat(rvt): add support for Risc-V tensor instructions

sploving1:
This commit introduces support for RISC-V tensor (RVT) extension (Internel Spec) instructions in NEMU, implemented via a new CONFIG_CUSTOM_TENSOR compile-time configuration option.

Key changes:

Added a new configuration flag CONFIG_CUSTOM_TENSOR to enable/disable RVT extension support

Implemented instruction decoding and execution logic for relevant RVT instructions

Extended the CPU state structure to include necessary tensor registers and state when enabled

The implementation ensures:

No impact on existing functionality: When CONFIG_CUSTOM_TENSOR is disabled (default), all existing NEMU features remain unchanged with zero performance or behavior impact

Clean separation: All RVT-related code is conditionally compiled based on CONFIG_CUSTOM_TENSOR

Maintainability: The changes follow existing NEMU architectural patterns and coding standards

The RVT extension support is disabled by default and can be enabled  by directly setting CONFIG_CUSTOM_TENSOR=y in the configuration. This allows users to experiment with tensor instructions while maintaining full compatibility with standard RISC-V operation.


poemonsense (commented):
If you need us to test this config, please add a config file to `configs` (e.g., `xs-ref-rvt_defconfig`). CI will automatically test its compilation.

include/cpu/decode.h
```diff
@@ -189,6 +201,16 @@ static inline void pattern_decode(const char *str, int len,
 #define def_INSTR_TABW(pattern, tab, width) def_INSTR_IDTABW(pattern, empty, tab, width)
 #define def_INSTR_TAB(pattern, tab)         def_INSTR_IDTABW(pattern, empty, tab, 0)
 
+#ifdef CONFIG_RVT
+#define def_INSTR_raw_64(pattern, body) do { \
+  if ((s->isa.instr.val >> 56) == pattern) { body; } \
```
cebarobot: Why do you need a special marco for 64-bit instruction? It seems that the origin ones work for 64-bit too. Is it because marco magic is too abstruct?

include/memory/paddr.h
```diff
@@ -42,6 +42,13 @@ uint8_t* guest_to_host(paddr_t paddr);
 /* convert the host virtual address in NEMU to guest physical address in the guest program */
 paddr_t host_to_guest(uint8_t *haddr);
 
+#ifdef CONFIG_RVT
+/* convert the uram guest physical address in the guest program to host virtual address in NEMU */
+uint8_t* uram_guest_to_host(paddr_t paddr);
```
cebarobot: What is a uram?

src/isa/riscv64/include/isa-all-instr.h
```diff
@@ -387,7 +400,8 @@
   ZFH_INSTR_TERNARY(f) \
   ZFA_INSTR_TERNARY(f) \
   ZFH_ZFA_INSTR_TERNARY(f) \
-  ZCB_INSTR_TERNARY(f)
+  ZCB_INSTR_TERNARY(f) \
+  TENSOR_INSTR(f)
```
cebarobot: Why here? Is tensor instr ternary?

sploving1:
> If you need us to test this config, please add a config file to `configs` (e.g., `xs-ref-rvt_defconfig`). CI will automatically test its compilation.

OK. I will add one such config.

src/isa/riscv64/include/isa-all-instr.h
```diff
@@ -171,6 +171,19 @@
...
 
+
+#ifdef CONFIG_RVT
+#define TENSOR_INSTR(f) \
+f(tx_matmul_subfloat) f(tx_matmul_float) f(tx_matmul_fix) f(tx_conv) f(tx_pooling) \
+f(tx_tcsr_xchg_nmask) f(tx_tcsr_xchg_mask) f(tx_tcsr_imm_nmask)  f(tx_tcsr_imm_mask) f(tx_tcsr_tshuffle) \
+f(tx_async_uram_copy) f(tx_async_uram_load)  f(tx_async_uram_store)  \
+f(tx_sync_uram_copy) f(tx_sync_uram_load)  f(tx_sync_uram_store)  \
+f(tx_int) f(tx_fp) f(tx_lut_int) f(tx_lut_fp) f(tx_int_reduce) f(tx_fp_reduce) \
+f(tx_convert) f(tx_sfu)
```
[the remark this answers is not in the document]
sploving1: Thanks.  I will change it！

src/isa/riscv64/instr/rvt/conv2d.c
```diff
@@ -0,0 +1,257 @@
...
+ * @param load_output Function to load existing output value (for accumulation)
+ * @param store_output Function to store result with appropriate data type
+ *
+ * @note All computations are performed in FP32 precision for accuracy
+ * @note No SIMD optimizations in this baseline implementation
+ * @note Padding handling: Only valid padding (no explicit zero padding buffer)
+ */
+static inline void conv2d_accum_kernel(
+    void *input, void *filter, void *output, int batch, int in_height,
+    int in_width, int in_channel, int filter_height, int filter_width,
```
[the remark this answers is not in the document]
sploving1: Thank you for the suggestion. Grouping related parameters into structs would greatly improve code readability and maintainability, and this refactoring is certainly worth doing. However, due to the amount of work involved and current time constraints, we plan to proceed with the merge as is and will schedule the structural improvements in a follow-up commit.

src/isa/riscv64/Kconfig
```diff
@@ -93,6 +93,11 @@ config RVV
   bool "RISC-V Vector Extension v1.0"
   default y
 
+config RVT
+  depends on !FPU_NONE
+  bool "RISC-V Tensor Extension v1.0"
```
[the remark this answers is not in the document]
sploving1: Done

include/cpu/decode.h
cebarobot: Have you checked the performance data for this? I don't think there will be huge difference in performance. 

The `pattern_decode` function uses some marco and compiler magic to most thing during preprocess and compilation. When running, it will be only a shift, a mask and a compare.

poemonsense:
> > Performance CI failed and said that `Error: Unhandled error: HttpError: Resource not accessible by integration`. Do you know why? I cannot tell from the outputs. @cyyself
> 
> I think that’s about privilege check in GitHub, because this PR comes from a personal fork instead of a branch of the NEMU repo. I will find a solution soon.

Thanks. Then I think we can merge this even if the performance comment is not ready now

src/isa/riscv64/instr/rvt/exec.h
```diff
@@ -0,0 +1,1784 @@
...
+ *PSL v2. You may obtain a copy of Mulan PSL v2 at:
+ *          http://license.coscl.org.cn/MulanPSL2
+ *
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY
+ *KIND, EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+ *NON-INFRINGEMENT, MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+ *
+ * See the Mulan PSL v2 for more details.
+ ***************************************************************************************/
+
```
cebarobot: `exec.h` is a subtle file in NEMU. It is codes that intend to be included in some places, rather than a normal header file. So there is a limit for it: you should not include any header files, or define functions. Let me explain it:

* When `PERF_OPT` is off, `rvt/exec.h` works as normal header file. `def_EHelper` expands to a function defination.
* When `PERF_OPT` is on, `rvt/exec.h` will be included via `isa-exec.h` into function `execute()` in file `cpu-exec.c`. `def_EHelper` expands to a label and some sentences.

If you defines some functions in the `exec.h`, you are defining functions in a function body, which is not good and annoys clang. (It's strange that gcc won't complain this).

You may use `make menuconfig` to change the compiler to clang. I advice you to fix the compilation failure under clang.

See #366.

sploving1: Thanks for pointing this. I fixed the clang compilation issue. As you explained, I removed function definitions in exec.h to ensure def_EHelper can be properly expanded under different compilation options.
The code now compiles correctly with both gcc and clang.

src/isa/riscv64/instr/decode.c
```diff
@@ -102,6 +105,94 @@ def_THelper(main) {
...
+
+  // Tensor SFU compute instructions.
+  def_INSTR_IDTAB(
+      "10010000????????????????????????????????????????????????????????",
+      tx_sfu, tx_sfu);
+
+  // Tenesor convert instructions.
+  def_INSTR_IDTAB(
+      "10001110????????????????????????????????????????????????????????",
+      tx_convert, tx_convert);
```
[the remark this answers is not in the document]
sploving1: Implemented as suggested. Now decoding uses the highest 1-2 bits to distinguish between 3R1W, 2R1W, and 1R1W formats, then dispatches to specific handlers in rvt/decode.h.

Key changes made:
1. Simplified instruction pattern matching using bitmasks as recommended
2. Reduced DHelper functions to match the three format categories
3. The code is now cleaner and more maintainable

Thank you for the guidance on this improvement.

cyyself:
Note: We have a new config in `configs/riscv64-xs-kunminghu-v3-ref_defconfig`, please don't forget add `# CONFIG_RVT is not set` in this config.

src/isa/riscv64/Kconfig
```diff
@@ -93,6 +93,11 @@ config RVV
   bool "RISC-V Vector Extension v1.0"
   default y
 
+config RVT
```
cyyself: A minor suggestion, since it's not a standard extension, I would like to suggest renaming `RVT` to `CUSTOM_TENSOR`.

src/isa/riscv64/instr/decode.c
```diff
@@ -102,6 +105,18 @@ def_THelper(main) {
   return table_inv(s);
 };
 
+#ifdef CONFIG_CUSTOM_TENSOR
+def_THelper(main_64) {
+  // 3R1W format instruction
+  def_INSTR_IDTAB("0???????????????????????????????????????????????????????????????", rvt_3r1w_dispatch, rvt_3r1w_dispatch);
```
cebarobot: If you rename the DHelper, it will be:
```
def_INSTR_IDTAB("0???????????????????????????????????????????????????????????????", rvt_3r1w, rvt_3r1w_dispatch);
```

PR #965, opened 2025-12-14
fix(ci): Separate PR comment workflow

cyyself:
When a PR is created from a forked repository, the CI workflow does not have permission to post comments back to the PR due to GitHub security restrictions. To address this, we separate the PR comment workflow from the main CI workflow. The main CI workflow runs with limited permissions, while the PR comment workflow runs with elevated permissions to post comments.


cyyself:
The `pr-comment.yml` can only be triggered after merged to master. I have tested this design here: https://github.com/cyyself/ci-perf-test/pull/1

PR #988, opened 2026-03-05
fix(mmu): adjusting the timing of non-virt PBMT assignments

Anzooooo:
We have already assigned a value to cpu.pbmt at the end of gpa_stage. It should not be reassigned, as this will cause it to be overwritten incorrectly:
https://github.com/OpenXiangShan/NEMU/blob/112f2fd1e68d1c8978f964f6a2846104dcd928e8/src/isa/riscv64/system/mmu.c#L371-L382


cebarobot (approved):
LGTM

PR #995, opened 2026-04-13
feat(bf16): add scalar and vector BF16 support

yu-yake2002:
Introduce Zfbfmin, Zvfbfmin, and Zvfbfwma support across decode and execution paths, including BF16 conversion/fused operations and related decode fixes for stable scalar/vector behavior.


yu-yake2002:
BF16 extensions will be used in kunminghu V2R2A and V3. @livypad has prepared some test cases, which will be merged into nexus-am in the near future.

Another issue that needs to be discussed is: Which configs in NEMU require enabling the three options of BF16? Next, we will update some configuration files based on the discussion results of the issue above.

yu-yake2002:
https://github.com/OpenXiangShan/nexus-am/pull/70 
There are some test cases. After enabling the BF16 options in NEMU, this program is able to run correctly.

src/isa/riscv64/instr/rvv/vcompute_impl.c
lewislzh: Has the NaN-boxing check here failed to take the BF16 format into account?

lewislzh, line 1198: Missing illegal-instruction checks for BF16 instructions, which only support the SEW=16 format.（only judge e16 e8,here bf16 not support e32, e64）

lewislzh (changes requested):

src/isa/riscv64/instr/rvv/vcompute_impl.c
yu-yake2002: The parameter `vsew` of `check_isFpCanonicalNAN` is indeed insufficient to distinguish between fp16 and bf16. I have replaced this parameter with `fp_type`.

yu-yake2002: Thanks. Added a vsew check for bf16 vector instructions.

lewislzh (approved):

yu-yake2002:
@good-circle This PR can be merged now. Let's see when to do it.

good-circle (approved):

PR #996, opened 2026-04-15
refactor(device): refactor uart16550 implementation

klin02:


poemonsense (commented):
I think we need to add a new 16550 model for uart, instead of replacing the legacy one

poemonsense (approved):
wait for @Tang-Haojin

Tang-Haojin (approved):

good-circle (approved):

PR #1005, opened 2026-04-22
fix(device): emulate dw-apb-uart probe registers

poemonsense:
Expand the uart16550 MMIO window to 0x1000 and add the minimal DesignWare extension registers required by the dw-apb-uart probe path, including USR, RE_EN, DLF, CPR, and UCV.

Keep the existing low-register 16550 behavior intact while avoiding offset aliasing between the classic register block and DesignWare extensions.


poemonsense:
@copilot review again

Tang-Haojin (approved):

good-circle (approved):

PR #1012, opened 2026-04-27
riscv64/mmu: predecode PMP and PMA entries

xiaokamikami:
PMP/PMA permission checks previously rebuilt TOR and NAPOT address metadata from CSRs on every physical-memory access.

Predecode the configuration, address bounds and NAPOT masks of active PMP/PMA entries during initialization and after architectural CSR writes. Refresh the derived state after full DiffTest CSR imports and retain a one-element backing array for supported zero-active-entry configurations.

Cover pmacfg0/2/4/6 and pmaaddr0-31 so high PMA entries receive the same WARL, locking and cache-refresh behavior as low entries. PMP priority, TOR chaining, partial-access rejection, privilege-mode handling and normal/Matrix access permissions remain unchanged.

The original xs-ref-noshare Linux/CoreMark measurement improves from 16,974,470 to 18,296,252 instr/s (+7.79%). The complete XS REF Shared Object suite reduces host instructions from 490,674,898,071 to 279,162,343,692 (-43.11%) with the same 85,270,182 guest instructions. Dynamic CSR writes, randomized permission checks, Linux boot and checkpoint restore all pass.


poemonsense (approved):

poemonsense (commented):
will pmp/pma CSRs be changed via csr_write?

Lemover (approved):
LGTM.

good-circle (approved):

PR #1022, opened 2026-05-09
fix(difftest): limit STORE_QUEUE_SIZE to avoid OOM

klin02:
When NEMU is enabled to check difftest store as REF, it will push store to a cpp queue, and pop it when compared with Difftest. However, sometimes when Difftest has not Store_check (such as BasicDiff in XS), NEMU will always push new store but not pop one, which enlarge mem to over 100G, and being killed for OOM.

This change set a limit size 65536 to store_queue size: NEMU will drop new store when queue.size > limit, and set a overflow flag, which will only be used when Difftest check:
1. When Difftest disable store check, the overflow will be ignored.
2. When Difftest enable store check, the overflow will trigger check failed, meaning there is more than 65536 store not be checked.


src/memory/paddr.c
```diff
@@ -670,6 +670,7 @@ int check_store_commit(uint64_t *addr, uint64_t *data, uint8_t *mask) {
       result = 1;
     }
   }
+  if (store_queue_overflow()) result = 1;
```
poemonsense: Check overflow at the beginning of this function. If overflow, print some error information.

In this way, only when the checker is called, there will be overflow information.

src/memory/store_queue_wrapper.cpp
```diff
@@ -23,12 +23,19 @@ void spec_store_log_stack_copy() { store_log_stack = spec_store_log_stack;}
...
 std::queue<store_commit_t> cpp_store_event_queue;
+bool store_queue_has_overflow;
 
 void store_queue_reset() {
   cpp_store_event_queue = {};
+  store_queue_has_overflow = false;
 }
 
 void store_queue_push(store_commit_t store_commit) {
+  if (cpp_store_event_queue.size() >= CONFIG_DIFFTEST_STORE_QUEUE_SIZE) {
```
poemonsense: First, check overflow before checking size:

If (not_overflow && size >= SIZE)

Second, avoid warning here because the store checker may not be enabled and we should avoid the warning here in such cases

poemonsense (approved):

Anzooooo (approved):

PR #1027, opened 2026-05-14
perf(difftest): lazily prepare CSR shadow for REF

xiaokamikami:
## Summary

This PR optimizes NEMU when used as a difftest REF by avoiding redundant CSR shadow preparation on every `DIFFTEST_TO_DUT` register copy.

Previously, `isa_difftest_regcpy(..., DIFFTEST_TO_DUT)` always called `csr_prepare()`, even when the mirrored CSR state had not changed. This is expensive in REF mode because `checkregs()` may request REF register state frequently, while most steps do not modify CSR state.

This patch adds a dirty flag for the RISC-V difftest CSR shadow:
- mark the CSR shadow dirty when CSR state can change, including CSR writes, interrupt/trap handling, MIP updates, counter overflow updates, and AIA-related external syncs;
- clear the dirty flag after `csr_prepare()` or after `DIFFTEST_TO_REF` writes the DUT state back into REF;
- skip `csr_prepare()` on `DIFFTEST_TO_DUT` when the CSR shadow is clean;
- avoid repeated external interrupt/AIA sync work when the incoming payload is unchanged.

`CONFIG_CLINT_LOCAL_TIMER_INTERRUPT` still keeps the old always-prepare behavior, since timer-related CSR state can change without an explicit CSR write.

## Performance

NEMU-only REF microbench results:

- XiangShan REF config: `42.44 ns/call -> 20.85 ns/call`
  - about `2.04x` faster
  - about `50.9%` lower call latency

- Nutshell REF config: `24.13 ns/call -> 14.32 ns/call`
  - about `1.69x` faster
  - about `40.7%` lower call latency

End-to-end XiangShan Verilator FPGA_SIM is dominated by DUT simulation time, so the REF-side speedup is mostly hidden there.

## Validation

- Built NEMU shared REF successfully with the RISC-V XS REF config.
- Ran XiangShan Verilator FPGA_SIM with NEMU as REF for 500k guest cycles.
- `microbench.bin` completed with `MicroBench PASS`.
- No difftest mismatch or assertion failure was observed.

### DIFF
--- a/src/cpu/difftest/ref.c
+++ b/src/cpu/difftest/ref.c
@@ -24,6 +24,14 @@
 #include <cpu/cpu.h>
 #include <difftest.h>
 
+#ifdef CONFIG_ISA_riscv64
+void csr_difftest_mark_dirty(void);
+#endif
+
+static inline void difftest_mark_csr_dirty(void) {
+  IFDEF(CONFIG_ISA_riscv64, csr_difftest_mark_dirty());
+}
+
 unsigned ref_hartid = 0;
 
 extern void load_flash_contents(const char *flash_img);
@@ -272,6 +280,7 @@ void difftest_raise_mhpmevent_overflow(uint64_t mhpmeventOverflowVec) {
 }
 
 void difftest_non_reg_interrupt_pending(void *nonRegInterruptPending) {
+  difftest_mark_csr_dirty();
   memcpy(&cpu.non_reg_interrupt_pending, nonRegInterruptPending, sizeof(struct NonRegInterruptPending));
   isa_update_mip(cpu.non_reg_interrupt_pending.lcofi_req);
 #ifdef CONFIG_RV_IMSIC
@@ -287,6 +296,7 @@ void difftest_non_reg_interrupt_pending(void *nonRegInterruptPending) {
 
 void difftest_interrupt_delegate(void *interruptDelegate) {
 #ifdef CONFIG_RV_IMSIC
+  difftest_mark_csr_dirty();
   memcpy(&cpu.interrupt_delegate, interruptDelegate, sizeof(struct InterruptDelegate));
 #endif // CONFIG_RV_IMSIC
 }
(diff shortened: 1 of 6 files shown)


src/cpu/difftest/ref.c
```diff
@@ -310,6 +326,10 @@ void difftest_update_vec_load_pmem() {
 
 void difftest_sync_aia(void *src) {
 #ifdef CONFIG_RV_IMSIC
+  if (memcmp(&cpu.fromaia, src, sizeof(struct FromAIA)) == 0) {
```
poemonsense: This is called only when aia should be synced. So no need to check here

src/cpu/difftest/ref.c
```diff
@@ -274,6 +286,10 @@ void difftest_non_reg_interrupt_pending(void *nonRegInterruptPending) {
 
 void difftest_interrupt_delegate(void *interruptDelegate) {
 #ifdef CONFIG_RV_IMSIC
+  if (memcmp(&cpu.interrupt_delegate, interruptDelegate, sizeof(struct InterruptDelegate)) == 0) {
```
poemonsense: This is called only when interrupt should be synced. So no need to check here

src/cpu/difftest/ref.c
```diff
@@ -259,6 +267,10 @@ void difftest_raise_mhpmevent_overflow(uint64_t mhpmeventOverflowVec) {
 }
 
 void difftest_non_reg_interrupt_pending(void *nonRegInterruptPending) {
+  if (memcmp(&cpu.non_reg_interrupt_pending, nonRegInterruptPending, sizeof(struct NonRegInterruptPending)) == 0) {
```
poemonsense: This is called only when interrupt should be synced. So no need to check here

poemonsense (commented):
what if exception happens? I don't see exception handler here

poemonsense (approved):

rezado (commented):
Review summary:

This PR makes REF-side CSR shadow preparation lazy for difftest. I audited the CSR fields used by csr_prepare() and found several non-csr_write() runtime update paths that also need dirty marking.

Missing dirty paths fixed:
- xRET updates to mstatus/vsstatus/hstatus
- FP dirty state and fcsr/fflags updates
- VP/RVV dirty state and vector CSR updates
- CLINT MMIO writes that may affect timer pending bits in mip
- direct mie updates in disable_time_intr()

CONFIG_TDATA1_ICOUNT can also update tdata1.icount, but it is disabled in the target difftest config.

https://github.com/OpenXiangShan/NEMU/tree/fix-difftest-csr-dirty-marking

poemonsense (approved):

PR #1028, opened 2026-05-15
feat(difftest): support checkpoint generation from SHARE REF

good-circle:


poemonsense:
Please separate different commits to multiple PRs to ease code review

PR #1029, opened 2026-05-15
perf(ci): include ref regcpy in shared-object benchmark

xiaokamikami:


tools/ref-so-runner/runner.c
```diff
@@ -59,20 +60,23 @@ typedef struct {
   const char *image_path;
   uint64_t load_addr;
   uint64_t max_instructions;
+  uint64_t regcpy_interval;
```
poemonsense: This argument is unnecessary. We can default it to `1`

tools/ref-so-runner/gen-xs-ref-perf-defconfig.sh
```diff
@@ -60,16 +60,6 @@ path = Path(sys.argv[1])
 text = path.read_text()
 
 updates = [
```
poemonsense, line 62: why do we change this

tools/ref-so-runner/runner.c
```diff
@@ -352,6 +379,12 @@ int main(int argc, char **argv) {
     api.exec(1);
     if (api.get_abs_instr_count == NULL) {
       guest_instructions += 1;
+    } else {
+      guest_instructions = api.get_abs_instr_count();
+    }
```
poemonsense, line 367: At every `exec`, we sync the registers

poemonsense (commented):
It's too complicated for testing the regcpy. I think simply copying the registers at every step is ok.

poemonsense (approved):

PR #1031, opened 2026-05-15
feat(checkpoint): trim compressed pmem dumps by resident pages

good-circle:
Use mincore to find the highest resident pmem page when writing compressed checkpoints, so checkpoint generation avoids dumping untouched mmap-backed memory. Query pmem in fixed-size chunks to keep the mincore result buffer bounded, and fall back to the current scan boundary when residency probing fails.

### DIFF
--- a/src/checkpoint/serializer.cpp
+++ b/src/checkpoint/serializer.cpp
@@ -21,6 +21,7 @@
 #include <checkpoint/cpt_env.h>
 #include <checkpoint/path_manager.h>
 #include <checkpoint/serializer.h>
+#include <cstdlib>
 #include <cstdio>
 #include <profiling/profiling_control.h>
 
@@ -39,6 +40,10 @@
 #include <fstream>
 #include <gcpt_restore/src/restore_rom_addr.h>
 #include <zstd.h>
+#ifdef CONFIG_MEM_COMPRESS
+#include <sys/mman.h>
+#include <unistd.h>
+#endif
 #ifdef CONFIG_LIBCHECKPOINT_RESTORER
 #include "pb.h"
 #include "pb_encode.h"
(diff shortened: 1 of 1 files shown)


src/checkpoint/serializer.cpp
```diff
@@ -39,6 +40,10 @@
 #include <fstream>
 #include <gcpt_restore/src/restore_rom_addr.h>
 #include <zstd.h>
+#if CONFIG_MSIZE > 0x200000000UL
```
poemonsense: Is this `if` required?

src/checkpoint/serializer.cpp
```diff
@@ -146,12 +151,65 @@ extern unsigned long MEMORY_SIZE;
 }
 
 #ifdef CONFIG_MEM_COMPRESS
+// Some configurations reserve a huge mmap-backed pmem range, while only the low
+// part is actually touched by the workload/restorer. Dumping the full configured
+// range would make checkpoint generation impractically slow and produce oversized
+// files, so for large pmem we trim the dump to the highest resident page.
+static size_t get_effective_pmem_size(uint8_t *pmem, size_t total_size) {
+#if CONFIG_MSIZE <= 0x200000000UL
```
poemonsense: I think we don't need to check whether the memory is larger than 8GB.

src/checkpoint/serializer.cpp
```diff
@@ -146,12 +151,65 @@ extern unsigned long MEMORY_SIZE;
...
+    if (mincore(pmem + this_start, this_chunk, vec) != 0) {
+      free(vec);
+      return offset;
+    }
+
+    for (size_t i = num_pages; i > 0; i--) {
+      if (vec[i - 1] & 1) {
+        size_t actual_size = this_start + i * (size_t)page_size;
+        if (actual_size < min_size) actual_size = min_size;
+        if (actual_size > total_size) actual_size = total_size;
```
poemonsense: Will this happen? `actual_size > total_size`

src/checkpoint/serializer.cpp
good-circle: Yes, respected leader, you are absolutely right

src/checkpoint/serializer.cpp
good-circle: Yes, respected leader, you are absolutely right

src/checkpoint/serializer.cpp
good-circle: Yes, respected leader, you are absolutely right. There won't be such a situation now.

poemonsense (approved):

PR #1035, opened 2026-05-16
feat(difftest): generate checkpoint on difftest mismatch

good-circle:
Expose a difftest-triggered checkpoint entry point for SHARE ref builds. Reuse serializer register and memory dumping for REF-side checkpoints, require an explicit base path in SHARE mode, and keep non-SHARE checkpoint paths available.

### DIFF
--- a/src/checkpoint/serializer.cpp
+++ b/src/checkpoint/serializer.cpp
@@ -21,8 +21,10 @@
 #include <checkpoint/cpt_env.h>
 #include <checkpoint/path_manager.h>
 #include <checkpoint/serializer.h>
+#include <cerrno>
 #include <cstdlib>
 #include <cstdio>
+#include <cstring>
 #include <profiling/profiling_control.h>
 
 #include "../isa/riscv64/local-include/csr.h"
@@ -39,8 +41,10 @@
 #include <fcntl.h>
 #include <fstream>
 #include <gcpt_restore/src/restore_rom_addr.h>
-#include <zstd.h>
 #ifdef CONFIG_MEM_COMPRESS
+#include <zstd.h>
+#endif
+#if defined(CONFIG_MEM_COMPRESS) || defined(CONFIG_SHARE)
 #include <sys/mman.h>
 #include <unistd.h>
 #endif
@@ -140,19 +144,15 @@ Serializer::Serializer() :
 
 extern "C" {
 uint8_t *get_pmem();
-word_t paddr_read(paddr_t addr, int len, int type, int trap_type, int mode, vaddr_t vaddr);
-uint8_t *guest_to_host(paddr_t paddr);
 #include <debug.h>
 #include <device/flash.h>
 #include <checkpoint/semantic_point.h>
-extern void log_buffer_flush();
-extern void log_file_flush();
 extern unsigned long MEMORY_SIZE;
 extern uint64_t clint_get_mtime();
 extern uint64_t clint_get_mtimecmp();
 }
 
-#ifdef CONFIG_MEM_COMPRESS
+#if defined(CONFIG_MEM_COMPRESS) || defined(CONFIG_SHARE)
 // Some configurations reserve a huge mmap-backed pmem range, while only the low
 // part is actually touched by the workload/restorer. Dumping the full configured
 // range would make checkpoint generation impractically slow and produce oversized
@@ -199,8 +199,8 @@ static size_t get_effective_pmem_size(uint8_t *pmem, size_t total_size) {
   return total_size;
 }
 
-void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t *flash_addr) {
-  // We must dump registers before memory to store them in the Generic Arch CPT
+void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t *flash_addr,
+                               const char *checkpoint_base_path) {
   assert(regDumped);
   assert(pmem_addr);
   const size_t PMEM_SIZE = get_effective_pmem_size(pmem_addr, (size_t)MEMORY_SIZE);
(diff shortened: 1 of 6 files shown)


good-circle:
Should rebase master after https://github.com/OpenXiangShan/NEMU/pull/1034 is merged

src/checkpoint/serializer.cpp
```diff
@@ -199,24 +199,30 @@ static size_t get_effective_pmem_size(uint8_t *pmem, size_t total_size) {
...
   string flash_file_path;
 #endif
 
   string base_file_path;
   string memory_file_path;
 
-  if (checkpoint_state == SimpointCheckpointing) {
+#ifdef CONFIG_SHARE
+  if (!checkpoint_base_path) {
+    xpanic("SHARE checkpoint serialization requires checkpoint_base_path.\n");
```
poemonsense, line 219: This check should be moved to `ref.c`

src/cpu/difftest/ref.c
```diff
@@ -405,3 +405,31 @@ void difftest_store_log_restore() {
 }
 #endif
 #endif // CONFIG_STORE_LOG
+
+#ifdef CONFIG_SHARE
```
poemonsense, line 409: Why do we check `CONFIG_SHARE` in ref.c?

src/cpu/difftest/ref.c
```diff
@@ -405,3 +405,31 @@ void difftest_store_log_restore() {
...
+extern void serialize_reg_to_mem();
+extern void serialize_pmem_to_file(const char *base_filepath);
```
poemonsense, line 411: I think we need a unified function called `serialize_checkpoint` or similar.

src/cpu/difftest/ref.c
```diff
@@ -405,3 +405,31 @@ void difftest_store_log_restore() {
...
+    printf("[difftest_trigger_checkpoint] ERROR: base filepath is NULL\n");
+    return;
+  }
+
+  if (!get_pmem()) {
+    printf("[difftest_trigger_checkpoint] ERROR: pmem is NULL\n");
+    return;
+  }
+
+  printf("[difftest_trigger_checkpoint] Writing checkpoint to %s_memory_.gz\n", base_filepath);
```
poemonsense, line 430: These `printf`'s should use Log. I remember Log will output a formatted string including the source code path

src/checkpoint/serializer.cpp
```diff
@@ -140,19 +144,15 @@ Serializer::Serializer() :
...
 #include <checkpoint/semantic_point.h>
-extern void log_buffer_flush();
-extern void log_file_flush();
 extern unsigned long MEMORY_SIZE;
 extern uint64_t clint_get_mtime();
 extern uint64_t clint_get_mtimecmp();
 }
 
-#ifdef CONFIG_MEM_COMPRESS
+#if defined(CONFIG_MEM_COMPRESS) || defined(CONFIG_SHARE)
```
poemonsense, line 155: Why CONFIG_SHARE here

src/checkpoint/serializer.cpp
```diff
@@ -39,8 +41,10 @@
 #include <fcntl.h>
 #include <fstream>
 #include <gcpt_restore/src/restore_rom_addr.h>
-#include <zstd.h>
 #ifdef CONFIG_MEM_COMPRESS
+#include <zstd.h>
+#endif
+#if defined(CONFIG_MEM_COMPRESS) || defined(CONFIG_SHARE)
```
poemonsense, line 47: Why CONFIG_SHARE here

src/checkpoint/serializer.cpp
```diff
@@ -199,24 +199,30 @@ static size_t get_effective_pmem_size(uint8_t *pmem, size_t total_size) {
   return total_size;
 }
 
-void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t *flash_addr) {
-  // We must dump registers before memory to store them in the Generic Arch CPT
```
poemonsense, line 203: Should this comment be removed?

src/checkpoint/serializer.cpp
```diff
@@ -365,7 +379,10 @@ void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t
     free(flash_compress_buffer);
 #endif
     free(memory_compress_buffer);
-
+#else
+    xpanic("ZSTD checkpoint format requires CONFIG_MEM_COMPRESS. "
+           "Please enable CONFIG_MEM_COMPRESS or use GZ format.\n");
```
poemonsense, line 384: The checkpoint format is currently checked in `monitor.c`. We should not check here again

In SHARE mode, maybe we need to determine the format in `ref.c`. For example, check if COMPRESS is enabled and assign the format accordingly.

src/checkpoint/serializer.cpp
```diff
@@ -374,12 +391,11 @@ void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t
...
 #else
-void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t *flash_addr) {}
+void Serializer::serializePMem(uint64_t inst_count, uint8_t *pmem_addr, uint8_t *flash_addr,
+                               const char *checkpoint_base_path) {}
 #endif
 
-#ifdef CONFIG_MEM_COMPRESS
-extern void csr_writeback();
-
+#if defined(CONFIG_MEM_COMPRESS) || defined(CONFIG_SHARE)
```
poemonsense, line 398: Why SHARE here

src/cpu/difftest/ref.c
```diff
@@ -405,3 +405,31 @@ void difftest_store_log_restore() {
 }
 #endif
 #endif // CONFIG_STORE_LOG
+
+#ifdef CONFIG_SHARE
+extern void serialize_reg_to_mem();
```
poemonsense, line 410: I find that this `serialize_reg_to_mem` function does not make sense. During checkpointing, any architectural state should not be changed. We must preserve the original NEMU states at any circumstances.

Therefore, we need to fix the original checkpointing functions before this PR.

poemonsense (commented):
We may still need separated PRs, like

1) Fix the wrong `serialize_reg_to_mem` function since the serializer should never change the architectural states of NEMU.

2) Add a unified checkpoint function that takes a checkpoint into a file.

3) Enable checkpointing functions in SHARE and add the new ref API to call it

PR #1036, opened 2026-05-18
feat(trigger): sdtrig support tdata3

wissygh:


sinceforYy (approved):

poemonsense (approved):
I don't know the detailed design of RISC-V sdtrig. But the code itself looks good.

PR #1040, opened 2026-05-25
fix(RVH): check HLVX final physical execute permission

fuhuakai:
The RISC-V HLVX instructions are loads for exception reporting, but their target address must also satisfy execute permission checks. NEMU already modeled this during page-table permission checks by using the global HLVX state when checking PTE permissions.

The original code has a bug in the final physical PMP and PMA checks. The translated HLVX access reached paddr_read as a normal read, so PMP and PMA checked only read permission. A region with R=1 and X=0 could be read by HLVX, which made the directed HLVX PMA test miss the expected load access fault.

Fix this by adding a MEM_TYPE_READ_EXEC access type for the final physical read. Keep the trap classification as read, require both read and execute permission in PMP and PMA checks.

This has been verified using the workload provided at https://github.com/OpenXiangShan/XiangShan/issues/5995. After the XiangShan cores and NEMU jointly fixed this bug, enabling the difftest can successfully pass the test.

### DIFF
--- a/include/memory/host-tlb.h
+++ b/include/memory/host-tlb.h
@@ -19,7 +19,7 @@
 #include <common.h>
 
 struct Decode;
-word_t hosttlb_read(struct Decode *s, vaddr_t vaddr, int len, int type);
+word_t hosttlb_read(struct Decode *s, vaddr_t vaddr, int len, int type, int trap_type);
 void hosttlb_read_matrix(struct Decode *s, vaddr_t vbase, vaddr_t stride,
                          int row, int column, int msew, bool transpose,
                          bool isacc, int mreg_id);
--- a/src/memory/host-tlb.c
+++ b/src/memory/host-tlb.c
@@ -74,14 +74,14 @@ static paddr_t va2pa(struct Decode *s, vaddr_t vaddr, int len, int type) {
 }
 
 __attribute__((noinline))
-static word_t hosttlb_read_slowpath(struct Decode *s, vaddr_t vaddr, int len, int type) {
-  paddr_t paddr = va2pa(s, vaddr, len, type);
-  word_t data = paddr_read(paddr, len, type, type, cpu.mode, vaddr);
+static word_t hosttlb_read_slowpath(struct Decode *s, vaddr_t vaddr, int len, int type, int trap_type) {
+  paddr_t paddr = va2pa(s, vaddr, len, trap_type);
+  word_t data = paddr_read(paddr, len, type, trap_type, cpu.mode, vaddr);
   if (
     MUXDEF(CONFIG_RV_MBMC, isa_bmc_check_permission(paddr, len, 0, 0), true) &&
     likely(in_pmem(paddr))
   ) {
-    HostTLBEntry *e = type == MEM_TYPE_IFETCH ?
+    HostTLBEntry *e = trap_type == MEM_TYPE_IFETCH ?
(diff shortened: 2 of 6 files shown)


src/memory/host-tlb.c
```diff
@@ -110,21 +111,25 @@ static void hosttlb_write_slowpath(struct Decode *s, vaddr_t vaddr, int len, wor
   }
 }
 
-word_t hosttlb_read(struct Decode *s, vaddr_t vaddr, int len, int type) {
+word_t hosttlb_read(struct Decode *s, vaddr_t vaddr, int len, int type, int paddr_type) {
   Logm("hosttlb_reading " FMT_WORD, vaddr);
+  if (paddr_type == MEM_TYPE_READ_EXEC) {
+    paddr_t paddr = va2pa(s, vaddr, len, type);
+    return paddr_read(paddr, len, paddr_type, type, cpu.mode, vaddr);
+  }
```
good-circle: Do we need a separate check here? If this is HLVX, `hld_st` will be true in `has_two_stage_translation()`, so it will not take the fast path anyway. Is it sufficient to just make sure `paddr_type` is passed correctly?

include/memory/host-tlb.h
```diff
@@ -19,7 +19,7 @@
 #include <common.h>
 
 struct Decode;
-word_t hosttlb_read(struct Decode *s, vaddr_t vaddr, int len, int type);
+word_t hosttlb_read(struct Decode *s, vaddr_t vaddr, int len, int type, int paddr_type);
```
good-circle: It still feels a bit hard to distinguish them this way. Would it make sense to rename them to `access_type` and `trap_type` for better clarity?

Also, I think we should add some comments somewhere to explicitly explain the difference between these two types. At the moment, is HLVX the only case where they differ, or are there other instructions that also use different semantics for access checking vs. trap classification?

fuhuakai: Good suggestion. I will try to distinguish these two types and add comments to explain clearly the differences and various uses of them.

src/memory/host-tlb.c
fuhuakai: Yes, this code is indeed redundant. HLVX will not take this path. I will streamline the code to ensure there are no unnecessary logical elements.

src/memory/paddr.c
```diff
@@ -268,13 +268,27 @@ bool check_paddr(paddr_t addr, int len, int type, int trap_type, int mode, vaddr
...
+//   - MEM_TYPE_IFETCH_READ: PTW reads a PTE on behalf of ifetch.
+//     type = IFETCH_READ (checks R), trap_type = IFETCH_READ (raises EX_IAF).
+//   - MEM_TYPE_WRITE_READ: PTW reads a PTE on behalf of store/AMO.
+//     type = WRITE_READ (checks R), trap_type = WRITE_READ (raises EX_SAF).
+//   - MEM_TYPE_READ_EXEC: HLVX load.
+//     type = READ_EXEC (checks R+X), trap_type = READ (raises EX_LAF).
+// PTW types encode both roles in one value (raise_read_access_fault recognizes
+// them directly). HLVX is the only case where type and trap_type are different
+// values, because MEM_TYPE_READ_EXEC encodes the access semantics while the
+// trap must still be a load fault.
```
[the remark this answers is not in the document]
fuhuakai: Fine. I'll chage my expression a little.

PR #1046, opened 2026-05-28
feat(mpt): add mpt v0.4 support to NEMU

HaoyuanLiuHZPH:
This PR adds MPT v0.4 functionality into the NEMU. Memory Protection Tables (MPT) is a RISC-V hardware extension that enforces fine‑grained access control on physical memory. It is designed for multi‑tenant and multi‑supervisor‑domain environments, where traditional isolation mechanisms (PMP, MMU) are insufficient. MPT allows a root domain security manager (RDSM) to define per‑domain access policies (read, write, execute) for arbitrary physical memory regions, using a multi‑level table structure stored in memory. It works alongside PMP and the virtual memory system: an access is allowed only if all three protection mechanisms permit it.
  - Add `CONFIG_RV_MPT_CHECK` and MPT-enabled defconfigs.
  - Add the `mmpt` CSR and preserve it in register snapshots.
  - Implement MPT table walking and permission checks.
  - Integrate MPT checks into physical memory access and page table walk paths.
  - Add safe MPTE reads through `mpt_paddr_read()`, bypassing recursive MPT checks while preserving PMP/PMA/MMIO validation.
  - Add `mfence` instruction support to flush translation state after MPT updates.
  - Flush TLB on `mmpt` writes

the mpt napot is not currently supported by NEMU, as it is not supported by the RTL either. We plan to add this functionality later.


Lemover:
Good job, LGTM.

src/memory/paddr.c
```diff
@@ -253,6 +253,20 @@ bool check_paddr(paddr_t addr, int len, int type, int trap_type, int mode, vaddr
     }
     return false;
   }
+
+  #ifdef CONFIG_RV_MPT_CHECK
```
fuhuakai, line 322: This wires the MPT permission check only into `check_paddr()`. However, with
`CONFIG_PERF_OPT`, normal translated accesses use the Host TLB path in `vaddr.c`. A
Host TLB miss will call `paddr_read()` / `paddr_write()` and therefore check MPT
once, but after the entry is filled, `hosttlb_read()` / `hosttlb_write()` hit paths
directly use `host_read()` / `host_write()` without calling `check_paddr()` again.

This means the cached Host TLB entry is not coherent with MPT state. In particular,
regarding invalidating Host TLB entries: the current CSR side effect only flushes on
`satp` writes.

Please confirm whether the issue described above actually exists. Should we make Host
TLB entries coherent with the MPT state? I am concerned that the current
implementation carries significant risk, so please evaluate and fix it.

src/cpu/cpu-exec.c
```diff
@@ -635,6 +641,9 @@ uint64_t lightqs_restore_reg_snapshot(uint64_t n) {
   #ifdef CONFIG_RV_MBMC
   cpu.mbmc = reg_ss.mbmc;
   #endif
+  #ifdef CONFIG_RV_MPT_CHECK
+  reg_ss.mmpt = cpu.mmpt;
```
fuhuakai: Do you mean "cpu.mmpt = reg_ss.mmpt;" here? Seem to be opposite.

src/memory/paddr.c
```diff
@@ -253,6 +253,20 @@ bool check_paddr(paddr_t addr, int len, int type, int trap_type, int mode, vaddr
...
+    if (!isa_mpt_check_permission(addr, len, type, mode)) {
+      if (trap_type == MEM_TYPE_WRITE) {
+        raise_access_fault(EX_SAF, vaddr);
+      } else {
+        raise_read_access_fault(trap_type, vaddr);
+        Log("isa mpt check failed, vaddr=" FMT_WORD ", paddr=" FMT_PADDR ", len=0x%x, type=0x%x, mode=0x%x",
+          vaddr, addr, len, type, mode);
```
fuhuakai: Here should first "Log" and then "raise_access_fault". Because if we raise a fault first, the execution flow will jump away and will not come back to execute the log output here.

src/memory/paddr.c
HaoyuanLiuHZPH: the TLB now flushes when mmpt csr changes.

src/cpu/cpu-exec.c
HaoyuanLiuHZPH: fixed

src/memory/paddr.c
HaoyuanLiuHZPH: fixed

fuhuakai:
lgtm

src/isa/riscv64/system/mmu.c
```diff
@@ -1111,6 +1132,103 @@ bool isa_bmc_check_permission(paddr_t addr, int len, int type, int out_mode) {
...
+    (addr >> 43) & 0x1ff    /* pn3 */
+  };
+  uint64_t range_off[4] = {
+    (addr >> 12) & 0xf,     /* range_off0 */
+    (addr >> 21) & 0xf,     /* range_off1 */
+    (addr >> 30) & 0xf,     /* range_off2 */
+    (addr >> 39) & 0xf      /* range_off3 */
+  };
+  int level;
+  if (mmpt->MODE >= 2)
```
good-circle: It seems that mmpt does not have dedicated WARL write logic, so is it possible that mmpt->MODE could be 3 in this case? If so, should Smmpt64 be used instead of Smmpt52? (even though we may not be prepared to implement Smmpt64)

src/isa/riscv64/system/mmu.c
```diff
@@ -1111,6 +1132,103 @@ bool isa_bmc_check_permission(paddr_t addr, int len, int type, int out_mode) {
...
+    uint64_t mpte_addr = mpt_base + (pn[level] << 3);
+    if (!isa_pmp_check_permission(mpte_addr, PTE_SIZE, MEM_TYPE_READ, MODE_S) ||
+      !isa_pma_check_permission(mpte_addr, PTE_SIZE, MEM_TYPE_READ)) {
+      Log("pmp or pma check failed when mpt walk with addr: %lx", mpte_addr);
+      return false;
+    }
+    mpte = host_read(guest_to_host(mpte_addr), 8);
+    bool v = (mpte & 0x1) != 0;
+    bool l = (mpte & 0x2) != 0;
+    bool res = ((mpte >> 2) & 0xff)==0 && ((mpte >> 62) & 0x1)==0;
```
good-circle, line 1332: Could you add some comments here? Has the case where napot / xwr does not comply with the manual specification been fully considered? Is requiring all bits [9:2] to be zero a bit too broad? (I haven't read the manual carefully, but the lack of comments here makes it a bit confusing)

src/isa/riscv64/system/mmu.c
```diff
@@ -1111,6 +1132,103 @@ bool isa_bmc_check_permission(paddr_t addr, int len, int type, int out_mode) {
 }
 #endif
 
+#ifdef CONFIG_RV_MPT_CHECK
+bool isa_mpt_check_permission(paddr_t addr, int len, int type, int out_mode) {
+  uint8_t mpt_perm = 0;
+
+  if (mmpt-> MODE == 0 || out_mode == MODE_M) {
```
good-circle: Should MPRV be considered here? (Refer to get_mprv() in the PMP check)

src/isa/riscv64/system/mmu.c
```diff
@@ -1111,6 +1132,103 @@ bool isa_bmc_check_permission(paddr_t addr, int len, int type, int out_mode) {
...
+      Logtr("MPT disabled addr:%lx mptmode %d,outmode %d\n", addr, mmpt-> MODE,out_mode);
+    return true;
+  }
+
+  Logtr("Mpt function called with addr:%lx mptmode %d,outmode %d\n", addr, mmpt-> MODE,out_mode);
+
+  word_t mpt_base = (mmpt->PPN) << 12;
+  uint64_t pn[4] = {
```
good-circle, line 1306: For the Smmpt43 case, if bits 43 and above are not all zero, should an access fault be reported?

src/isa/riscv64/system/mmu.c
```diff
@@ -1111,6 +1132,103 @@ bool isa_bmc_check_permission(paddr_t addr, int len, int type, int out_mode) {
...
+  }
+  uint64_t mpte;
+  while(1){
+    uint64_t mpte_addr = mpt_base + (pn[level] << 3);
+    if (!isa_pmp_check_permission(mpte_addr, PTE_SIZE, MEM_TYPE_READ, MODE_S) ||
+      !isa_pma_check_permission(mpte_addr, PTE_SIZE, MEM_TYPE_READ)) {
+      Log("pmp or pma check failed when mpt walk with addr: %lx", mpte_addr);
+      return false;
+    }
+    mpte = host_read(guest_to_host(mpte_addr), 8);
```
good-circle: Although continuing to call paddr_read would cause a loop re-entering the MPT check, directly using host_read(guest_to_host()) would completely bypass the pmem / mmio address range checks. From the perspective of your design intent here, is this acceptable?

src/isa/riscv64/system/mmu.c
```diff
@@ -1111,6 +1132,103 @@ bool isa_bmc_check_permission(paddr_t addr, int len, int type, int out_mode) {
...
+    bool v = (mpte & 0x1) != 0;
+    bool l = (mpte & 0x2) != 0;
+    bool res = ((mpte >> 2) & 0xff)==0 && ((mpte >> 62) & 0x1)==0;
+    bool res_l= ((mpte >> 58) & 0x1f)==0;
+    uint64_t data = ((mpte << 1) >> 1) >>10;
```
good-circle, line 1334: Should napot be considered here, or do we not support it?

src/isa/riscv64/system/mmu.c
```diff
@@ -303,10 +303,20 @@ paddr_t gpa_stage(paddr_t gpaddr, vaddr_t vaddr, int type, int trap_type, bool i
...
+    bool check_failed;
+
+    #ifdef CONFIG_RV_MPT_CHECK
+      check_failed = !isa_pmp_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ, MODE_S) ||
+        !isa_pma_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ) ||
+        !isa_mpt_check_permission(p_pte, PTE_SIZE,MEM_TYPE_READ,1);   
+    #else
+      check_failed = !isa_pmp_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ, MODE_S) ||
+        !isa_pma_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ);
+      Log("pmp or pma check failed when PTW");
```
good-circle: If we disable CONFIG_RV_MPT_CHECK, does that mean line 315 will always print this log?

src/isa/riscv64/reg.c
```diff
@@ -216,6 +216,13 @@ void isa_reg_display() {
      DISPLAY_CSR("mbmc", mbmc->val);
      printf("\n");
    #endif
+   
+  #ifdef CONFIG_RV_MPT_CHECK
+    DISPLAY_CSR("mmpt", mmpt->val);
+    printf("\n");
+  #else
+    printf("MPT disabled!\n");
```
good-circle: This should behave like other extensions, only print output if enabled, remain silent if disabled. There is no need to additionally print that MPT is not enabled.

good-circle (commented):
Sorry for the late review. I have a few more review comments. Please take a moment to confirm them when you have time.

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: no, we still use 52.mmpt should be WARL

HaoyuanLiuHZPH: the default mode is smmpt43 in rtl instead of 52. I will change accrodingly

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: the napot is not currently supported by the rtl, will be add later

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: fixed

HaoyuanLiuHZPH: also for smmpt52

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: yes，fixed

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: we dont support it

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: fixed

src/isa/riscv64/system/mmu.c
HaoyuanLiuHZPH: fixed

src/isa/riscv64/reg.c
HaoyuanLiuHZPH: get it

good-circle (approved):

yu-yake2002 (approved):

src/isa/riscv64/system/mmu.c
```diff
@@ -596,9 +606,20 @@ static paddr_t ptw(vaddr_t vaddr, int type) {
...
+    bool check_failed;
+
+    #ifdef CONFIG_RV_MPT_CHECK
+      check_failed = !isa_pmp_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ, MODE_S) ||
+        !isa_pma_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ) ||
+        !isa_mpt_check_permission(p_pte, PTE_SIZE,MEM_TYPE_READ,1);
+    #else
+      check_failed = !isa_pmp_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ, MODE_S) ||
+        !isa_pma_check_permission(p_pte, PTE_SIZE, MEM_TYPE_READ);
+      Log("pmp or pma check failed when PTW");
```
huxuan0307, line 618: WTF?

PR #1053, opened 2026-06-08
perf(cpu-exec): skip timer in ref/shared mode

klin02:
`cpu_exec` uses get_time() only for standalone NEMU runtime statistics, including host elapsed time and MIPS reporting. The value is not part of REF architectural state when NEMU is built as a DiffTest shared library.

Skip this accounting under CONFIG_SHARE to avoid repeated host time queries. In the measured run, get_time accounted for 20% speedup in riscv64-nemu-interpreter-so for fpga difftest.

### DIFF
--- a/src/cpu/cpu-exec.c
+++ b/src/cpu/cpu-exec.c
@@ -781,7 +781,9 @@ void cpu_exec(uint64_t n) {
     Loge("Setting NEMU state to RUNNING");
   }
 
+#ifndef CONFIG_SHARE
   uint64_t timer_start = get_time();
+#endif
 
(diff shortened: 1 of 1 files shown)


poemonsense (approved):

poemonsense:
Maybe we should check where g_timer is used

klin02:
> Maybe we should check where g_timer is used

Search in NEMU repo, g_timer is only used in final perf report:
https://github.com/OpenXiangShan/NEMU/blob/40daca91c7aefc9411842d0f6fe3d5dae1682517/src/cpu/cpu-exec.c#L155-L162

PR #1055, opened 2026-06-08
perf(riscv64/aia): speed up priority index lookup

klin02:
AIA priority handling maps IRQ numbers back to priority-group indices. The previous helper scanned interrupt_default_prio[] on each query, and high_iprio() repeated the xei lookup in the loop.

Build a reverse lookup table from interrupt_default_prio[] during AIA priority initialization, and cache xei_prio_idx for repeated high_iprio()


klin02:
CI performance does not change because the CI workload does not include the AIA query path. 
Tested on FPGA/Geekbench and works well.

poemonsense (approved):

PR #1058, opened 2026-06-08
feat(monitor): provide bootloader rng-seed via FDT

klin02:
Linux can block while collecting entropy when the boot path does not provide a trusted seed source. This can make system-mode workloads hang around early random initialization.

Patch the loaded DTB before guest boot to provide /chosen/rng-seed and append random.trust_bootloader=on to bootargs.

The seed is read from host /dev/urandom, so Linux can initialize its CRNG without changing the workload image or DTS source. Enable the feature for riscv64-xs_defconfig.


poemonsense (commented):
Should we do this in NEMU?

poemonsense (approved):

PR #1060, opened 2026-06-09
fix(rvv decode): add inst[14:12] constraint on vlm/vsm

skyhgzsh:


poemonsense (approved):

poemonsense (commented):
I don't know the vector load/store instructions and should not approve this

PR #1071, opened 2026-06-14
feat(checkpoint): enable gzip images by default

good-circle:
Gzip image loading and checkpoint serialization only require zlib, which is already linked by both normal and SHARE builds. Do not tie the GZ paths to CONFIG_MEM_COMPRESS.

Keep ZSTD support behind CONFIG_MEM_COMPRESS because that option still controls linking libzstd.


src/checkpoint/serializer.cpp
```diff
@@ -148,7 +152,7 @@ extern uint64_t clint_get_mtime();
 extern uint64_t clint_get_mtimecmp();
 }
 
-#ifdef CONFIG_MEM_COMPRESS
+#if defined(CONFIG_MEM_COMPRESS) || defined(CONFIG_SHARE)
```
poemonsense: I don't think it's good to use CONFIG_SHARE here. We should try to avoid using this macro in non-REF code.

Why do we need this? should COMPRESS be enabled if we need checkpoint in so?

src/checkpoint/serializer.cpp
```diff
@@ -676,4 +687,10 @@ void serialize_reg_to_mem() {
   serializer.serializeRegs(get_pmem());
 }
 
+#ifdef CONFIG_SHARE
```
poemonsense: Is this ifdef really needed?

poemonsense (commented):
We should be careful when using CONFIG_SHARE. Avoid using it wherever possible

src/checkpoint/serializer.cpp
good-circle: Yes, dear senior brother, `gz` no longer depends on `COMPRESS`, so we do not need to enable `COMPRESS` under `SHARE`. I have removed the corresponding `#ifdef`.

src/checkpoint/serializer.cpp
good-circle: Yes, dear senior brother, we do not really need it here

poemonsense (commented):
`CONFIG_MEM_COMPRESS` was used to enable/disable gz lib in unsupported platforms. If we need to enable it by default, maybe we should remove this configuration?

good-circle:
> `CONFIG_MEM_COMPRESS` was used to enable/disable gz lib in unsupported platforms. If we need to enable it by default, maybe we should remove this configuration?

Originally, when gzip support was added, `CONFIG_MEM_COMPRESS` was only used to guard the gzip path and to decide whether `-lz` should be added during linking.

Later, `-lz` became enabled by both DUT NEMU and REF NEMU builds. In the current XiangShan configs, and also in our CentOS/Ubuntu environments, zlib is available and supported. Meanwhile, zstd support was added later, and `CONFIG_MEM_COMPRESS` now mainly controls whether we compile zstd code and link `-lzstd`.

So maybe the config name is no longer accurate. Would it be better to rename it to something like `CONFIG_ZSTD_COMPRESS`?

poemonsense:
> So maybe the config name is no longer accurate. Would it be better to rename it to something like `CONFIG_ZSTD_COMPRESS`?

Yes

poemonsense (approved):

PR #1072, opened 2026-06-14
feat(checkpoint): expose explicit checkpoint control for SHARE

good-circle:
Expose C entry points so SHARE users can request checkpoint serialization with a caller-provided output prefix and select the flash checkpoint layout explicitly.

### DIFF
--- a/src/checkpoint/serializer.cpp
+++ b/src/checkpoint/serializer.cpp
@@ -504,6 +504,10 @@ void Serializer::serialize(uint64_t inst_count, const char *checkpoint_base_path
 #endif
 }
 
+void Serializer::setStoreCptInFlash(bool enable) {
+  store_cpt_in_flash = enable;
+}
+
 #ifdef CONFIG_LIBCHECKPOINT_RESTORER
 
 #define USING_METADATA_OVERRIDE_DEFAULT_MEMLAYOUT(variable_name) \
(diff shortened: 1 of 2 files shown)


src/cpu/difftest/ref.c
```diff
@@ -405,3 +406,17 @@ void difftest_store_log_restore() {
 }
 #endif
 #endif // CONFIG_STORE_LOG
+
+#ifdef CONFIG_SHARE
```
poemonsense: Why is the macro used here?

ref.c is used only when SHARE

good-circle: Sorry, dear senior brother, you are right

src/checkpoint/serializer.cpp
```diff
@@ -481,6 +481,10 @@ void Serializer::serialize(uint64_t inst_count, const char *checkpoint_base_path
   __attribute__((unused))
   uint64_t serialize_buffer_size;
 
+  if (checkpoint_base_path) {
+    store_cpt_in_flash = true;
```
poemonsense: Why should we use this `if`?

good-circle: Previously I assumed that an explicit checkpoint_base_path would only be used by the difftest-triggered checkpoint path. Since that path also needs to force the checkpoint into flash, I tied these two conditions together.

But they are actually unrelated: checkpoint_base_path should only describe the output prefix, while whether to use the flash checkpoint layout should be controlled separately. I have updated this so the difftest trigger explicitly enables flash checkpointing before calling serialize_checkpoint().

poemonsense (approved):
Maybe we need to update the PR/commit title/message when merging

PR #1074, opened 2026-06-15
perf(riscv64): fast path interrupt polling

xiaokamikami:
Optimize the hot RISC-V interrupt polling path in `isa_query_intr()`.

Most polling calls do not result in a taken interrupt. Avoid constructing the pending interrupt state when no interrupt is enabled, M-mode interrupts are globally disabled, or NMIE is closed.

After observing a pending interrupt, compute the effective privilege and delegation mask once. Use a static byte-sized priority table to reduce the overhead of the cold interrupt selection path.

The RISC-V interrupt priority, privilege, delegation, RVH, and Sscofpmf semantics are preserved.

DynamoRIO shared REF CoreMark results with identical guest instruction counts:

```text
Before: 6,466,360,574 host instructions
After:  5,507,075,491 host instructions
Reduction: 14.835%

### DIFF
--- a/src/isa/riscv64/system/intr.c
+++ b/src/isa/riscv64/system/intr.c
@@ -360,7 +360,7 @@ word_t isa_query_intr() {
     global_enabled_intr = 0;
     if (cpu.v && (((cpu.mode == MODE_S) && vsstatus->sie) || (cpu.mode < MODE_S))) global_enabled_intr |= deleg & hdeleg;
     if (((cpu.mode == MODE_S) && mstatus->sie) || (cpu.mode < MODE_S) || cpu.v) global_enabled_intr |= deleg & ~hdeleg;
-    if (((cpu.mode == MODE_M) && mstatus->mie) || (cpu.mode < MODE_M)) global_enabled_intr |= ~deleg;
+    if (cpu.mode < MODE_M) global_enabled_intr |= ~deleg;
   }
 #else
   word_t deleg = mideleg->val;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/isa/riscv64/system/intr.c [hunk 1]
```diff
@@ -340,50 +340,57 @@ word_t raise_intr(word_t NO, vaddr_t epc) {
...
+  word_t deleg = mideleg->val;
+  word_t global_enabled_intr;
+  if (!cpu.v && cpu.mode == MODE_M) {
+    global_enabled_intr = ~deleg;
+  } else {
+    word_t hdeleg = get_hideleg();
+    global_enabled_intr = 0;
+    if (cpu.v && (((cpu.mode == MODE_S) && vsstatus->sie) || (cpu.mode < MODE_S))) global_enabled_intr |= deleg & hdeleg;
+    if (((cpu.mode == MODE_S) && mstatus->sie) || (cpu.mode < MODE_S) || cpu.v) global_enabled_intr |= deleg & ~hdeleg;
+    if (((cpu.mode == MODE_M) && mstatus->mie) || (cpu.mode < MODE_M)) global_enabled_intr |= ~deleg;
```
sinceforYy: Is the condition “(cpu.mode == MODE_M) && mstatus->mie” a duplicate of the one on line 354?

sinceforYy (approved):

poemonsense (approved):

PR #1083, opened 2026-06-18
feat(difftest): add checkpoint trigger entry

good-circle:
Add difftest_trigger_checkpoint() so REF users can request checkpoint generation from NEMU.

Skip M-mode because the current checkpoint restore path does not reliably support restoring M-mode state. For supported modes, enable the flash checkpoint layout before serializing the checkpoint.

### DIFF
--- a/src/cpu/difftest/ref.c
+++ b/src/cpu/difftest/ref.c
@@ -405,3 +405,16 @@ void difftest_store_log_restore() {
 }
 #endif
 #endif // CONFIG_STORE_LOG
+
+extern void set_store_cpt_in_flash(bool enable);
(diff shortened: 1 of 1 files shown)


poemonsense (commented):
Why was there no compilation error before?

good-circle:
> Why was there no compilation error before?

PR8 starts using `load_img()` from `flash.c` via `memory/image_loader.h`. Without this fix, `flash.c` passes a `const char *` to the old `char *` prototype, which fails under `-Werror`. The implementation already uses `const char *`; https://github.com/OpenXiangShan/NEMU/pull/1085 just syncs the header before PR8 depends on it.

poemonsense (approved):

PR #1089, opened 2026-06-22
perf(riscv64): static inline PMP/PMA CSR accessors

xiaokamikami:
PMP/PMA CSR accessor helpers are on the hot path of memory permission checks. Keeping these small helpers out of line introduces repeated function-call overhead, although they only perform CSR array indexing or mask generation.

Move the PMP/PMA configuration, address, and TOR-mask accessors into csr.h as static inline helpers. This allows the compiler to fold the indexing and mask operations into the permission-check path without changing PMP/PMA state, CSR layout, or RISC-V permission semantics.

With riscv64-xs-ref-perf_defconfig and PERF_OPT=OFF, host instructions are reduced as follows:

- microbench: 752,246,127 -> 540,143,051 (-28.20%)
- coremark: 1,303,041,641 -> 985,279,499 (-24.39%)
- combined: 2,055,287,768 -> 1,525,422,550 (-25.78%)

The REF shared-object smoke test and the non-SHARE riscv64-xs_defconfig build both pass.

### DIFF
--- a/src/isa/riscv64/local-include/csr.h
+++ b/src/isa/riscv64/local-include/csr.h
@@ -2003,13 +2003,56 @@ void update_vstopi();
 #endif
 
 /** PMP **/
-uint8_t pmpcfg_from_index(int idx);
-word_t pmpaddr_from_index(int idx);
-word_t pmp_tor_mask();
+extern rtlreg_t csr_array[4096];
+
+#ifdef CONFIG_RV_PMP_CSR
+// get 8-bit config of one PMP entries by index.
+static inline uint8_t pmpcfg_from_index(int idx) {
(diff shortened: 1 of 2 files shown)


klin02:
Note the performance benefits comes from `static inline` . @xiaokamikami Please help provide corresponding PR title and final commit message.

poemonsense (approved):

klin02:
The performance benefits from `inline`, but inline declaration in priv.c is unseen from mmu.c when `CONFIG_CC_LTO` is not set. 
Maybe we can enable `CONFIG_CC_LTO`, and keep these function as its original position with only `inline` declaration. 
I will try in another PR.

klin02:
Implemented by #1113 . So close this PR.

PR #1091, opened 2026-06-24
ci(perf): add AME workload to correctness smoke

xiaokamikami:
AME perf workload is now checked in the earlier correctness stage before performance collection.

This catches AME-specific regressions earlier and keeps the perf benchmark tied to a known-good functional baseline, while preserving the existing diff-based correctness flow.

No perf measurement semantics are changed.


yu-yake2002:
We haven't added AME support for spike yet, so differential testing might not be possible here.

PR #1102, opened 2026-07-08
fix(RVH): use effective virtual mode for VA width check

fuhuakai:
RVH data accesses with MPRV can use MPP and MPV instead of the current virtualization state. The early virtual-address width check still used cpu.v when it selected the VS/G-stage address width.

This can reject a valid guest Sv48 data address as an invalid Sv39 address before page-table walking runs.

Use the effective virtual mode for non-instruction accesses when selecting whether VS/G-stage Sv39 or Sv48 translation is enabled. Keep instruction fetches based on cpu.v, since MPRV does not affect fetch.

### DIFF
--- a/src/isa/riscv64/system/mmu.c
+++ b/src/isa/riscv64/system/mmu.c
@@ -638,9 +638,9 @@ int isa_mmu_check(vaddr_t vaddr, int len, int type) {
   // Instruction fetch addresses and load and store effective addresses,
   // which are 64 bits, must have bits 63–39 all equal to bit 38, or else a page-fault exception will occur.
 #ifdef CONFIG_RVH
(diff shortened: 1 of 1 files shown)


fuhuakai:
A bug reported from the sting difftest.

Anzooooo (approved):

Anzooooo:
@fuhuakai Can they be merged?

fuhuakai:
@Anzooooo  I think they can. should not cause more bug.

Anzooooo:
Please merge one copy into kunminghu-v2.

fuhuakai:
@Anzooooo 
I have already submitted the exact same PR to the kunminghu-v2 branch of NEMU. Please approve it.
https://github.com/OpenXiangShan/NEMU/pull/1129

PR #1106, opened 2026-07-08
 feat: add optional instruction category statistics

rezado:
This PR adds an optional instruction-category statistics feature to NEMU. When enabled, the simulator records per-instruction execution IDs, aggregates counts by category (alu, load, store, branch, jump, csr_system, fp, vector, amo, fence, other), and prints a summary in monitor statistics at the end of execution. The implementation is guarded by config flags, so there is no behavior change when the feature is disabled, which is default.

Usage:

Enter menuconfig and enable Print instruction category statistics in Miscellanous.


src/cpu/instr_stat.c
```diff
@@ -0,0 +1,231 @@
...
+  }
+  instr_stat_exec_counts[exec_id]++;
+}
+
+#define INSTR_STAT_COUNT(name) + instr_stat_exec_counts[EXEC_ID_##name]
+#define INSTR_STAT_SUM(list) (0 list(INSTR_STAT_COUNT))
+
+#if defined(CONFIG_ISA_riscv64)
+
+#define RV64_ALU_INSTR(f) \
```
poemonsense, line 52: I'm wondering whether we really need this. This looks like aggregating the results again after decoding. Can we count the instructions during the decoder?

src/cpu/cpu-exec.c
```diff
@@ -468,6 +474,7 @@ static void execute(int n) {
     is_ctrl = false;
     Logti("prev pc = 0x%lx, pc = 0x%lx", prev_s->pc, s->pc);
 
+    IFDEF(CONFIG_INSTR_CATEGORY_STAT, instr_stat_count(this_s->exec_id));
```
poemonsense: Should we use `CONFIG_INSTR_CNT_BY_CATEGORY` instead of `CONFIG_INSTR_CNT_BY_INSTR`?

src/cpu/cpu-exec.c
```diff
@@ -741,6 +750,7 @@ void fetch_decode(Decode *s, vaddr_t pc) {
   s->snpc = pc;
   IFDEF(CONFIG_DEBUG, log_bytebuf[0] = '\0');
   int idx = isa_fetch_decode(s);
+  s->exec_id = idx;
```
poemonsense: 1. Is this really needed?
2. If so, we should wrap it with the macro (updating it only when the config is enabled)

src/cpu/instr_stat.c
rezado: Updated. The exec ID is now mapped to a category once during decoding, while the category counter is incremented when the instruction completes execution. Counting in the decoder would count decode/cache-fill events rather than dynamic instruction executions when tcache reuses a decoded block.

src/cpu/cpu-exec.c
rezado: Yes, it has been updated.

src/cpu/cpu-exec.c
rezado: 1. It has been changed to an uint8 InstrStatCategory.
2. Yes, it has been updated.

poemonsense (approved):

PR #1109, opened 2026-07-09
perf(ame): reduce matrix register access overhead

yu-yake2002:
- Gate hot AME matrix register access assertions behind AME_REG_ACCESS_CHECK.
- Disable AME_REG_ACCESS_CHECK for all AME defconfigs.
- Split whole-matrix load/store handling.


yu-yake2002:
Our group should prepare a CI test case suitable for evaluating the performance of AME in NEMU.

poemonsense (approved):
Don't know AME but the code itself looks good to me. Please feel free to merge

PR #1114, opened 2026-07-14
perf(compile): add configurable GCC aggressive-inline tuning

eastonman:
Adds a `CC_AGGRESSIVE_INLINE` Kconfig option (default `y`, `depends on CC_GCC`) that appends two GCC-specific tuning flags:

  - `--param max-inline-insns-single=256`
  - `-falign-labels=32:9:64:15`

and marks the interpreter's address-translation hot path `get_effective_address` (both `CONFIG_RVH` variants) and `isa_mmu_check` as `inline`.
The option is enabled across all GCC defconfigs; Clang configs keep out at the config level.

### Flag details

  - `--param max-inline-insns-single=256` — raises GCC's size limit for inlining a single function, so the small MMU/translation helpers get inlined
  into their callers instead of staying out-of-line.
  - `-falign-labels=n:m:n2:m2` aligns branch targets to an `n`-byte boundary while skipping at most `m-1` padding bytes, otherwise to an `n2`-byte
  boundary skipping at most `m2-1`. The value `32:9:64:15` therefore means: align a branch target to a **32-byte** boundary when it costs **≤8**
  padding bytes, else to a **64-byte** boundary when it costs **≤14** — keeping hot targets inside a single instruction-fetch window without
  over-padding cold code.

This reworks #747. Inlining the MMU/translation helpers and raising GCC's inlining threshold measurably speeds up the standalone interpreter.

## Benchmarks 

Standalone interpreter (`riscv64-xs-cpt`) built at the current `-O3 + LTO` ceiling, with vs. without this change. Host: AMD EPYC 9684X (Zen 4).
  Metric: NEMU's reported `simulation frequency`; SPEC CPU2006 checkpoints; median of 5 CPU-pinned runs.

  | workload   |   Δ    |
  |------------|--------|
  | h264ref    | +13.0% |
  | gcc        | +11.4% |
  | bzip2      |  +9.2% |
  | gobmk      |  +5.9% |
  | libquantum |  +5.2% |
  | mcf        |  +2.4% |
  | lbm        |  +0.8% |
  | milc       |  +0.7% |
  | namd       |  +0.4% |
  | bwaves     |  -2.3% |
  | **geomean**| **+4.5%** |


poemonsense (approved):
LGTM. Let's wait for performance CI results

eastonman:
It increases the dynamic instruction counts though.

poemonsense:
> It increases the dynamic instruction counts though.

So dynamic instruction count increases but performance increases?

This is shown in the `NEMU Throughput` column. I think so

eastonman:
Yeah I agree, the -falign param allows more aggresive label alignment, which pads hot loops with nop.

PR #1119, opened 2026-07-15
feat(matrix): Support XSAI AME proposal 14

ecall73:
Require whole-register load/store encodings to use rs2=x0.

Derive fixed row strides internally for contiguous whole-register transfers.


yu-yake2002 (approved):

configs/riscv64-matrix-xs-dual-ref_defconfig
yu-yake2002, line 1: You don't need to roll back these two configuration files. After I updated difftest, XSAI was already aligned with NEMU.

configs/riscv64-matrix-xs-ref_defconfig
yu-yake2002, line 1: You don't need to roll back these two configuration files. After I updated difftest, XSAI was already aligned with NEMU.

yu-yake2002 (changes requested):

yu-yake2002 (approved):

PR #1122, opened 2026-07-20
feat(COVE, CFI): add CFI landing pad support to NEMU

HaoyuanLiuHZPH:
This PR adds configurable RV64 Zicfilp forward-edge CFI support to NEMU.
- Add `CONFIG_RV_ZICFILP` and an XS defconfig with the extension enabled.
- Track Expected Landing Pad (ELP) state for `JALR`, `C.JR`, `C.JALR`, and `ret`.
- Validate `LPAD` targets during instruction execution through LPAD instruction, including alignment and `x7` label checks.
- Add Zicfilp CSR fields and controls.
- Preserve and restore ELP across traps, `MRET`, `SRET`, and `MNRET`.
- Raise software-check exceptions for invalid landing-pad targets.
Zicfiss shadow-stack support are planned for a follow-up.


src/isa/riscv64/instr/rvi/compute.h
```diff
@@ -90,6 +90,26 @@ def_EHelper(andi) {
...
+    if (rd != 0) {
+      riscv64_raise_software_check(LANDING_PAD_FAULT);
+    }
+    if ((s->pc & 0x3) != 0) {
+      riscv64_raise_software_check(LANDING_PAD_FAULT);
+    }
+    uint32_t lpl = (s->isa.instr.val >> 12) & 0xFFFFF;
+    uint32_t x7_lpl = (reg_l(7) >> 12) & 0xFFFFF;
+
+    printf("[ZICFILP-DEBUG] Checking lpad at pc=0x%lx: instr=0x%08x, expected_lpl(x7)=0x%x, actual_lpl=0x%x\n", s->pc, s->isa.instr.val, x7_lpl, lpl);
```
Lemover: Debug log need to be removed. If it is essential, "printf" can be replaced with Log, or something  else.

src/isa/riscv64/instr/rvc/exec.h
```diff
@@ -50,6 +65,22 @@ def_EHelper(c_jr) {
 
 def_EHelper(c_jalr) {
   rtl_li(s, &cpu.gpr[1]._64, s->snpc);
+
+#ifdef CONFIG_RV_ZICFILP
```
Lemover: The Ziclip code block appears in many instructions' EHelper. They seems to the same with each other.  May be we should wrapper it with a function and call the function.

Lemover:
Other modificatins LGTM.

.gitignore
```diff
@@ -26,5 +26,6 @@ __pycache__/
 include/config
 include/generated
 build
+tmp
```
yu-yake2002: Which piece of code will generate the `tmp` directory, or generate files under the `tmp` directory?

src/isa/riscv64/instr/rvc/exec.h
```diff
@@ -37,6 +37,21 @@ def_EHelper(c_j) {
...
 
 def_EHelper(c_jr) {
+#ifdef CONFIG_RV_ZICFILP
+  // Zicfilp: check if landing pad is expected after indirect jump
+  {
+    bool zicfilp_en = riscv64_zicfilp_enabled(cpu.mode, MUXDEF(CONFIG_RVH, cpu.v, false));
+    if (zicfilp_en) {
+      uint32_t rs1 = (s->isa.instr.val >> 15) & 0x1f;
+      if (rs1 != 1 && rs1 != 5 && rs1 != 7) {
+        cpu.elp = 1;
```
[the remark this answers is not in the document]
HaoyuanLiuHZPH: fixed

src/isa/riscv64/instr/pseudo.h
```diff
@@ -65,6 +65,21 @@ def_EHelper(p_jal) {
...
 def_EHelper(p_ret) {
+#ifdef CONFIG_RV_ZICFILP
+  // Zicfilp: check if landing pad is expected after indirect jump
+  {
+    bool zicfilp_en = riscv64_zicfilp_enabled(cpu.mode, MUXDEF(CONFIG_RVH, cpu.v, false));
+    if (zicfilp_en) {
+      uint32_t rs1 = (s->isa.instr.val >> 15) & 0x1f;
+      if (rs1 != 1 && rs1 != 5 && rs1 != 7) {
+        cpu.elp = 1;
+      } else {
```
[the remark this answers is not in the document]
HaoyuanLiuHZPH: not a bug

src/isa/riscv64/instr/rvi/compute.h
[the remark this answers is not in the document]
HaoyuanLiuHZPH: fixed

.gitignore
HaoyuanLiuHZPH: I'll have it removed

src/isa/riscv64/instr/rvc/exec.h
HaoyuanLiuHZPH: EHelper is now wrapped into a function and called at C_jr C_jalr JALR

yu-yake2002 (approved):
I am not familiar with CFI, but it does not appear to affect the functionality or performance of AME.

src/cpu/cpu-exec.c
```diff
@@ -437,6 +440,16 @@ static void execute(int n) {
     __attribute__((unused)) rtlreg_t ls0, ls1, ls2;
     br_taken = false;
 
+#ifdef CONFIG_RV_ZICFILP
+    if (unlikely(cpu.elp == 1)) {
+      uint32_t target_instr = vaddr_ifetch(s->pc, 4);
```
good-circle: I think the unconditional `vaddr_ifetch(pc, 4)` may introduce a spurious instruction page fault.

There are three cases, for example:

1. The target instruction is actually 32-bit. In this case NEMU must fetch all 4 bytes anyway. If fetching the upper 16 bits raises an instruction page/access fault, that fault is legitimate and has higher priority than the landing-pad software check.

2. The target instruction is 16-bit, so after fetching the first 16 bits we already know it cannot be an LPAD. If the extra upper 16-bit fetch does not fault, the final result is still the expected landing-pad software-check exception.

3. The problematic case is when the target instruction is 16-bit, but the extra upper 16 bits cross a page boundary and fault. Those bytes are not part of the current instruction and should never have been fetched. However, `vaddr_ifetch(pc, 4)` will propagate that instruction page/access fault instead of raising the expected landing-pad software-check exception.

So maybe the LPAD check should preserve NEMU's normal variable-length fetch behavior, rather than unconditionally fetching 4 bytes?

src/cpu/cpu-exec.c
```diff
@@ -736,6 +749,17 @@ static void execute(int n) {
     cpu.debug.current_pc = s.pc;
     cpu.pc = s.snpc;
     ref_log_cpu("pc = 0x%lx inst %x", s.pc, s.isa.instr.val);
+
+#ifdef CONFIG_RV_ZICFILP
+    if (unlikely(cpu.elp == 1)) {
+      uint32_t target_instr = vaddr_ifetch(s.pc, 4);
```
good-circle: same as line R445

good-circle (commented):
I don't have any other comments. Could @Lemover please take another look? If you think it's ok, let's merge it into master(

good-circle (approved):

src/cpu/cpu-exec.c
HaoyuanLiuHZPH: fixed

Lemover (approved):

good-circle:
@HaoyuanLiuHZPH The PR description here doesn't need to follow your description above. It still needs to describe in more detail what exactly was changed (e.g., the xxx module, added xxx functionality, added xxx support, etc.). For things that are not yet supported and are planned for the next phase, you can also note them as well (referring to your previous https://github.com/OpenXiangShan/NEMU/pull/1046 would be fine).

PR #1134, opened 2026-07-30
perf(riscv64): skip inactive mcontrol6 checks at call sites

xiaokamikami:
Hot execute, scalar load/store, AMO, CBO, and vector load/store paths enter the full mcontrol6 checker even when the trigger array contains no active mcontrol6 entry.

Check the authoritative trigger types at those call sites and skip the checker in the common inactive case. CSR writes and DiffTest imports update the same trigger array, so no derived state or synchronization is needed.

Active-trigger access order, timing, exception handling, and vector restart behavior are unchanged.


poemonsense (approved):
LGTM. Can we further cache the trigger mcontrol6 state?

xiaokamikami:
> LGTM. Can we further cache the trigger mcontrol6 state?

Thanks! Let's merge #1134 first. I will then rebase the mcontrol6 state-cache follow-up onto the latest master and use the cached state in the call-site predicate.

PR #1136, opened 2026-08-05
perf(share): trim single-step execution boundary

xiaokamikami:
This optimization targets one-instruction shared REF difftest execution. Standard SHARE_REF configurations do not use autonomous CLINT interrupts.

### DIFF
--- a/src/cpu/cpu-exec.c
+++ b/src/cpu/cpu-exec.c
@@ -151,6 +151,19 @@ static void update_instr_cnt() {
 #endif // CONFIG_ENABLE_INSTR_CNT
 }
 
+static inline void update_instr_cnt_after_execute() {
+#if defined(CONFIG_SHARE) && !defined(CONFIG_LIGHTQS) && \
+    defined(CONFIG_INSTR_CNT_BY_INSTR)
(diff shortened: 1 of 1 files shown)


poemonsense (commented):
The idea looks good. But are the macro used correctly?

xiaokamikami:
> The idea looks good. But are the macro used correctly?

Thanks for pointing this out. CONFIG_SHARE_REF was being used as a proxy for several execution properties, which made the conditions broader than necessary.
I updated the PR to use the actual properties instead:
single-step count settlement is guarded by the shared single-step and instruction-counting constraints;
the post-instruction interrupt query is skipped only for a one-instruction batch, since the outer execution boundary polls again before another instruction can execute;
timer refreshes are compiled according to CONFIG_CLINT_LOCAL_TIMER_INTERRUPT.
I also removed the redundant PERF_OPT check, since PERF_OPT already depends on !SHARE. LightQS, non-deterministic REF, and local-CLINT configurations have been rebuilt, and the directed MTIP boundary test passes.

poemonsense (approved):

PR #1142, opened 2026-08-11
Gem5 ref enable tval

BirdsTweetWindWhisper:


src/isa/riscv64/init.c
```diff
@@ -69,7 +69,7 @@ void init_isa() {
   // For RV64 systems, if U-mode is not supported, then UXL is hardwired to zero.
   mstatus->val = 0xaUL << 32;
   // initialize the value fs and vs to 0
-  mstatus->fs = 1;
+  mstatus->fs = 0;
```
poemonsense, line 72: Why this

poemonsense (commented):
Please separate different patches to different PRs so that we can review a single change at one time

PR #1146, opened 2026-08-13
feat(matrix): support XSAI AME proposal-14 e4 load/store

ecall73:


include/ame/raw-fp4.h
yu-yake2002, line 1: The license for this code should include BOSC.

src/isa/riscv64/ame/raw_fp4_memory.c
yu-yake2002, line 1: This file should have a license at the top.

include/ame/mstore.h
yu-yake2002, line 1: Is `raw_fp4` necessary? Can we use `msew = 7` to represent an element width of 4 bits?

yu-yake2002 (changes requested):

include/ame/mstore_queue_wrapper.h
```diff
@@ -32,4 +33,3 @@ bool mstore_queue_check_addr_conflict(uint64_t addr, int len);
 #endif
 
 #endif // __AME_MSTORE_QUEUE_WRAPPER_H__
-
```
yu-yake2002: The file should be guaranteed to end with a newline character.

src/isa/riscv64/instr/ame/mcompute.h
```diff
@@ -97,7 +97,9 @@ bool is_signed_int_mtype(uint64_t type_code) {
 def_EHelper(mmacc) {
   // Check MS
   require_matrix();
+#ifndef CONFIG_RV_AME_FP4
```
yu-yake2002: What is the meaning here? If we enable FP4 support, do we not need to consider the mp bit?

src/isa/riscv64/instr/ame/mldst.h
```diff
@@ -24,6 +24,9 @@
 #include "../local-include/intr.h"
 #include "../local-include/rtl.h"
 #include "../local-include/reg.h"
+#ifdef CONFIG_RV_AME_FP4
+#include <ame/raw-fp4.h>
```
yu-yake2002: Why are `<>` used here instead of `""`)?

yu-yake2002 (changes requested):

yu-yake2002 (approved):

poemonsense (approved):

PR #1147, opened 2026-08-14
perf(rvv): bulk copy unmasked direct VLE

xiaokamikami:
Use a bulk copy for eligible unmasked unit-stride VLEs in MMU_DIRECT mode.

The fast path is limited to single-field, same-page PMEM accesses after trigger, PMP, PMA, and MBMC checks. Sparse memory and all unsupported cases continue to use the original path.

### DIFF
--- a/src/isa/riscv64/instr/rvv/vldst_impl.c
+++ b/src/isa/riscv64/instr/rvv/vldst_impl.c
@@ -302,39 +302,61 @@ void vld(Decode *s, int mode, int mmu_mode) {
 
   bool fast_vle = false;
 
-#if !defined(CONFIG_SHARE) && !defined(CONFIG_TDATA1_MCONTROL6)
+#if !defined(CONFIG_SHARE)
   uint64_t start_addr = base_addr + (vstart->val * nf) * s->v_width;
   uint64_t last_addr = base_addr + (vl_val * nf - 1) * s->v_width;
   uint64_t vle_size = last_addr - start_addr + s->v_width;
   __attribute__((unused)) bool cross_page = last_addr / PAGE_SIZE != start_addr / PAGE_SIZE;
+#ifndef CONFIG_TDATA1_MCONTROL6
   uint8_t masks[VLMAX_8] = {0};
+#endif
 
   Logm("vld start_addr: %#lx, v_width: %u, vl_val: %lu, vle size=%lu, vstart->val: %lu, nf=%lu",
       base_addr, s->v_width, vl_val, vle_size, vstart->val, nf);
 
-  if (is_unit_stride && nf == 1 && vl_val > vstart->val && vtype->vlmul < 4 && !cross_page) {
-    s->last_access_host_addr = NULL;
-    extern void dummy_vaddr_data_read(struct Decode *s, vaddr_t addr, int len, int mmu_mode);
-    dummy_vaddr_data_read(s, start_addr, s->v_width, mmu_mode);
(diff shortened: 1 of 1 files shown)


src/isa/riscv64/instr/rvv/vldst_impl.c
```diff
@@ -299,17 +386,33 @@ void vld(Decode *s, int mode, int mmu_mode) {
...
+      isa_pma_check_permission(start_addr, vle_size, MEM_TYPE_READ)) {
+    isa_vec_misalign_data_addr_check(start_addr, s->v_width, MEM_TYPE_READ);
+    void *reg_file_addr = NULL;
+    get_vreg_with_addr(vd, vstart->val, &tmp_reg[1], eew, 0, 0, 0, &reg_file_addr);
+    Assert(reg_file_addr != NULL, "reg_file_addr is NULL");
+    memcpy(reg_file_addr, guest_to_host(start_addr), vle_size);
+    fast_vle = true;
+  }
+#ifndef CONFIG_TDATA1_MCONTROL6
+  else if (is_unit_stride && nf == 1 && vl_val > vstart->val && vtype->vlmul < 4 && !cross_page) {
```
huxuan0307: I think that nesting the fast path condition will be more readable.
The fast path condition should not contain the decode condition.
```c++
if (decode condition) {
  if (fast path condition) {
    // fast path
  } else {
    // origin slow path
  } 
}
```

src/isa/riscv64/instr/rvv/vldst_impl.c
```diff
@@ -412,19 +515,27 @@ void vld(Decode *s, int mode, int mmu_mode) {
...
+  // Keep the generic tail path for shapes that cannot use the contiguous fill.
+  bool lmul8_tail_fast = false;
+#if !defined(CONFIG_MULTICORE_DIFF)
+  lmul8_tail_fast = RVV_AGNOSTIC && mode == MODE_UNIT && s->v_nf == 0 &&
+      eew == 0 && vemul == 3 && vtype->vta && vl_val > ori_vstart;
+#endif
+  if (lmul8_tail_fast) {
+    // For EEW=8, EMUL=8 the eight aligned vector registers are contiguous.
+    memset(cpu.vr[vd]._8 + vl_val, 0xff, VLMAX_8 - vl_val);
+  } else if (RVV_AGNOSTIC && (mode == MODE_MASK || vtype->vta) && vl_val > ori_vstart) {   // set tail of vector register to 1
```
huxuan0307: This fast pash should also be nested in the body of decode condition, and the duplicated condition should be removed.

huxuan0307 (changes requested):

src/isa/riscv64/instr/rvv/vldst_impl.c
xiaokamikami: Thanks. Updated as suggested. The common decode eligibility checks are now in the outer condition, while the direct VLE fast path only contains its additional runtime checks. The original fallback path is nested under the same decode condition.

src/isa/riscv64/instr/rvv/vldst_impl.c
xiaokamikami: Thanks. Updated as suggested. The LMUL=8 tail fast path is now nested under its decode condition, and the duplicated lmul8_tail_fast condition has been removed. The generic tail path remains in the else branch.

PR #1159, opened 2026-08-18
ci(perf): benchmark XSAI BLAS workloads in NEMU

yu-yake2002:
Add the pinned xsai-blas v1.0.0 workload archive to the performance job, cache and unpack it, and run it with riscv64-matrix-xs_defconfig.

Add an xsai-blas suite to perf-benchmark.sh with the three release workloads while preserving the existing default suite and baseline comparison flow.


yu-yake2002:
The current AME performance testing takes a relatively long time, but I am using the AVX instruction set to optimize AME simulation, and have achieved significant improvements. A pull request will be submitted soon.

poemonsense (approved):

PR #1179, opened 2026-08-25
fix(ame): skip disabled matrix store tracking

yu-yake2002:


poemonsense (approved):
please feel free to merge the PR as this is for AME

PR #1199, opened 2026-09-07
fix(cpu-exec): preserve current PC across tcache flush

ha0lyu:
Under `CONFIG_PERF_OPT`, `prev_s` records the instruction currently being executed. However, `rtl_priv_next()` calls `tcache_handle_flush()` before jumping to `end_of_loop`.

The tcache flush path calls `save_globals(ex.tnext)`, which updates `prev_s` to the next instruction too early. As a result, the trace incorrectly reports the next instruction as the previous instruction.

For example, executing the instruction at `0x800002c4` produced:
```
  [execute] prev pc = 0x800002bc, pc = 0x800002c0
  [execute] end_of_loop: prev pc = 0x800002c4, pc = 0x800002c4
```
The end_of_loop message should report:
```
  [execute] end_of_loop: prev pc = 0x800002c0, pc = 0x800002c4
```

### DIFF
--- a/src/cpu/cpu-exec.c
+++ b/src/cpu/cpu-exec.c
@@ -278,9 +278,11 @@ static bool manual_cpt_quit = false;
   do {                                                                         \
     if (g_sys_state_flag) {                                                    \
       if (g_sys_state_flag & SYS_STATE_FLUSH_TCACHE) {                         \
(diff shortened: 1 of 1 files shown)


poemonsense (approved):

poemonsense:
I'm wondering whether this affects performance

PR #1200, opened 2026-09-07
fix(ame): enforce MMACC architectural semantics

yu-yake2002:
Validate MMACC operand encodings, tile bounds, and supported integer and floating-point type combinations.


src/engine/interpreter/fp.c
```diff
@@ -480,7 +480,6 @@ def_rtl(mfpcall, rtlreg_t *dest, const rtlreg_t *src1, const rtlreg_t *src2, uin
 #ifndef CONFIG_FPU_NONE
   uint32_t w = FPCALL_W(cmd);
   uint32_t op = FPCALL_OP(cmd);
-  isa_fp_csr_check();
```
poemonsense, line 483: why this

yu-yake2002: This code is used for matrix floating-point operations. The previous implementation was incorrect — `isa_fp_csr_check` checks the `FS` field, but the matrix extension does not depend on the value of `FS`.

poemonsense (approved):

PR #1207, opened 2026-09-10
configs: Add validated GEM5 CI reference variants

jensen-yan:
## Motivation

GEM5 CI needs reproducible references built from one NEMU revision, while retaining distinct scalar/RVV, H and multicore semantics. Sharing a source revision does not mean sharing one `.so`.

Companion: https://github.com/OpenXiangShan/GEM5/pull/1105

## Changes

- Add four fully expanded configurations: `riscv64-gem5-ci-{normal,normal-dedup,h,multi}-ref_defconfig`, matching the repository's `bump_configs.sh` convention.
- Keep existing `gem5-ref` and `gem5-multicore-ref` defconfigs unchanged.
- All four resolve to Sv48 and the same 1376-byte RVV/RVH/VCSR/FCSR register layout, with trigger-register comparison disabled and guided TVAL enabled.
- Normal variants preserve the extensions needed by scalar CI, including Zacas for GCC16 SPEC17. H retains undisturbed vector behavior; multi retains the multicore protocol. Neither inherits the scalar extension superset, especially SSTC.
- Document memory pairing, build selection and consumer ABI requirements concisely in `docs/gem5_ref.md`.

This is configuration-only. It does not claim complete RVA23 support (Zabha and Smcdeleg are not enabled), change NEMU source behavior, or overwrite deployed shared references.

## Validation

Base is `d30fff1ece9e0480146caf504660a0239b994eef`, also the revision used by the companion GEM5 release. All four new defconfigs were expanded with that revision's native Kconfig tool and compared byte-for-byte against the previously built and tested `.config` files: all identical.

The corresponding four libraries were built successfully and each exports `DIFFTEST_REG_SIZE=1376`. With the companion GEM5 changes:

- Full vector microtests: **878/878 passed**.
- Focused SPEC06 GCC12/GCC16, SPEC17 GCC16 AMOCAS.D and SPEC26 checkpoint tests passed.
- H checkpoint/restorer, SMT CoreMark/mcf and true two-core CHI tests passed.
- Four final-build 1M-instruction normal/dedup A/B pairs had identical simulated instructions, ticks, cycles and IPC.

These results use the equivalent configurations before this configuration-only commit, not a newly executed full regression at the PR SHA. A fresh isolated `make` encountered dependency-download/ccache environment restrictions; configuration equivalence was verified directly using the already-built same-revision Kconfig tool. Full SPEC/H/SMT coverage remains outstanding. Binary hashes, pinned dependencies and detailed boundaries are recorded in the companion GEM5 PR's `util/nemu_ref/VALIDATION.md`.


poemonsense (commented):
NEMU docs should be put in `docs` directory. docs should be simple and avoid unnecessary words.

jensen-yan:
Moved the documentation to docs/gem5_ref.md and shortened it to build/selection instructions and the ABI/memory constraints. Also fixed Check All Configs: the initial savedefconfig outputs were semantically correct but did not satisfy bump_configs.sh, which requires fully expanded configurations. All four now match their Kconfig-generated .config byte-for-byte, with no effective configuration changes. The published GEM5 reference release is unchanged.

poemonsense (approved):

eastonman:
Finally. Great.

PR #1210, opened 2026-09-14
fix(config): suppress clang-18+ warning

ngc7331:
Intel avx10.1 dropped optional 512-bit in the new [whitepaper](https://cdrdv2.intel.com/v1/dl/getContent/784343)

With `march=native`, when some CPU reports cpuid with `avx10.1-256`, clang will promote it to `avx10.1-512` and issue a warning. And we have `-Werror` so the compile fails.

It seems that on GitHub CI, most of the runners uses older CPU and does not report `avx10.1-256` in cpuid. So this only happens occasionally.

This commit suppresses that warning.

Link: https://github.com/actions/runner-images/issues/14296
Link: https://en.wikipedia.org/wiki/CPUID


poemonsense (approved):

eastonman:
Jesus christ. GitHub is using AVX10.1 machines now? Amazing.

PR #1212, opened 2026-09-14
feat(ame): check scalar/vector store visibility before matrix loads

yu-yake2002:
## Summary

The existing checker diagnoses matrix-store → scalar/vector-load conflicts, but does not detect matrix loads consuming scalar/vector stores without the required synchronization.

This PR adds the missing direction and renames `AME_MSTORE_ACCESS_CHECK` to `AME_MEM_ACCESS_CHECK`. The option remains disabled by default, and the existing matrix-store checks are preserved.

- Track completed scalar/vector RAM stores in a `svstore` queue and warn when a matrix load overlaps stores not yet covered by a supported visibility barrier.
- Recognize `mfence`, qualifying normal FENCE and acquire-release atomic operations, and completed same-block CBO clean/flush operations. Failed SC/AMOCAS operations do not publish stores; partial CBO coverage retains the uncovered bytes.
- Treat PBMT=IO RAM stores as belonging to both W and O fence domains.
- Add hooks for physical-memory and HostTLB accesses, bulk vector stores, and raw FP4 matrix loads.
- Check REF matrix loads in guest instruction order, before deferred AMU execution, so later stores or fences cannot change an earlier load's diagnosis.
- Keep diagnostics warning-only, reporting producer/consumer PCs, virtual and physical addresses, and access sizes. `mfence` does not replace the existing `mrelease`/`macquire` protocol for the opposite direction.

### Performance

Measured eight runs per configuration after one warmup each, alternating configurations and pinning execution to the same CPU. Host: AMD EPYC 9684X, GCC 13.3.0; CI optimization settings, including LTO and `-march=x86-64-v3`, were retained.

| Median metric | Checker off | Checker on |
| --- | ---: | ---: |
| NEMU host execution time | 1.581 s | 1.887 s |
| Guest throughput | 77.88 M instr/s | 65.27 M instr/s |
| Peak RSS | 12 MiB | 144 MiB |

Enabling both checking directions increased execution time by **19.3%**, reduced throughput by **16.2%**, and added **132 MiB** of peak RSS for this workload. This measures the combined checker overhead, not the new `svstore` tracking in isolation.

All measured runs reached GOOD TRAP with identical guest instruction counts and no warnings in either direction. The kernel contains `fence rw, rw` before matrix loads and `mrelease`/`macquire` after matrix stores, consistent with the current checker policy. No issue was detected under that policy; this is not a proof of complete memory-model correctness.


yu-yake2002:
@Gs-ygc This feature has already been implemented, and the performance overhead is within an acceptable range. XSAI-BLAS provides three test cases for NEMU, and the test here uses the one with the medium scale.

poemonsense (approved):

PR #1216, opened 2026-09-16
fix: change log format in raise_intr function

ha0lyu:
This PR will fix "%lx" print `NO = 0x8000000000000009` as `-9223372036854775799`, which is not convenient to debug.


ha0lyu:
How can I commit into master and kunminghu-v2 branch at the same time?

poemonsense (approved):

